PR #85, opened 2013-10-19
Various stuff

learn-more:
Fix crashing on invalid playerid (32)
Update logging timestamp (don't really need the Dayname in it)
Add compiler info to window title
msvc fixes (linking related, less runtime dependencies)
add commandline connect option (also protocol)
Highlight the bbox that would be hit (only on teammates, for debugging the hittest)
Centerprint player's killer


feikname:
Whoah there, almost used all the tags on this pull.

PR #340, opened 2015-01-24
New try to minimaps improvements

pandaro:
Edit:
the DevPaks27.zip is now in my dropbox, all works fine, but, please feel free to bring back DevPaks.zip in the old places

### DIFF
--- a/Sources/Client/ScoreboardView.cpp
+++ b/Sources/Client/ScoreboardView.cpp
@@ -303,7 +303,7 @@ namespace spades {
 					IntVector3 Colorplayer=IntVector3::Make(palette[i][0],palette[i][1],palette[i][2]);
 					Vector4 ColorplayerF = ModifyColor(Colorplayer);
 					ColorplayerF *=1.0f;
-					font->Draw(buf, MakeVector2(colX + 35.f - size.x,rowY),1.f, changeF);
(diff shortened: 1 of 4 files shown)


Sources/Client/ScoreboardView.cpp
```diff
@@ -241,6 +257,50 @@ namespace spades {
...
+						{128,255,255},//24 light azure
+						{255,255,128},//25 light yellow
+						{255,128,255},//26 light pink2
+						
+						{165,42,42},//27 brown
+						{255,69,0},//28 orange	red
+						{255,165,0},//29  orange
+						
+						{139,69,19},//30 maroon medium
+						{210,105,30},//31 choccolate
```
theunknownxy: This table is exactly the same as in MapView.cpp. I think this should be deduplicated.

pandaro: yes i know, but i dont know  whether it is better position to put it in: MapView.cpp or ScoreboardView.cpp.
Second question. This requires something in particular about the place, or the syntax, where it needs to be declared the color palette?

pandaro: think it is appropriate to define the table within the namespace MapView.h openspades?
something like this:
http://pastebin.com/xhFB0H1C

pandaro: Ok now there is only one definition of  the array palette.
other issues?

pandaro:
No one have other issues?

PR #342, opened 2015-02-04
Update copyright holder to "OpenSpades Developers"

yvt:


Sources/AngelScript/source/as_symboltable.h
```diff
@@ -1,6 +1,6 @@
 /*
    AngelCode Scripting Library
-   Copyright (c) 2012 Andreas Jonsson
```
theunknownxy: I'm not a lawyer, but I think that you shouldn't change this copyright notice.

yvt: Thanks for noticing.

Issue #314, opened 2014-08-25
Users can't fire their weapons at all when sprint is held


SK83RJOSH:
This doesn't quite make sense, the expected behavior (and the behavior in the official client) would be that you can fire so long as your character isn't actually sprinting. It's really confusing to not be able to fire despite the fact that my weapon is it's ready position. This is particularly problematic when you crouch and happen to be holding sprint for instance, because you make have crouched momentarily to take a shot.

yvt:
Fixed by #384.

PR #375, opened 2015-12-12
#314 Pressing sprint key stops you from using weapon/tool 

someoneigna:
See (#314)

### DIFF
--- a/Sources/Client/Player.cpp
+++ b/Sources/Client/Player.cpp
@@ -118,11 +118,7 @@ namespace spades {
 			
 			if(!IsAlive())
 				return;
-			
-			if(input.sprint){
-				newInput.primary = false;
-				newInput.secondary = false;
-			}
+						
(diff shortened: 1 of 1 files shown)


Kurtoid:
:+1:

Chameleonhider:
Did you check compatibility with netcode? I believe this would conflict, as client still sends server your sprint state and server doesn't care if you actually sprint or not.

Instead, I would like to suggest doing this in Client_Input.cpp:
if (input.crouch || (!input.moveBackwards && !input.moveForwards && !input.moveLeft && !input.moveRight))
{
  input.sprint = false
}

So as a result, if you are crouching or not pressing WASD keys, sprint key is ignored; you can shoot, aim weapon and server does not receive that you are pressing down sprint key.

yvt:
As #384 fixed this issue, rejecting this pull request.

PR #384, opened 2016-02-10
Configure-able player name drawing

Chameleonhider:
Currently, hottracked player names are drawn above crosshair together with distance to them. Center of screen is important region and thus it should be configure-able.  This allows you to disable drawing name or just disable drawing the distance. Also possible to position the name wherever you want. Currently position of drawn name is not checked for sake of code's simplicity.

Please see http://buildandshoot.com/viewtopic.php?f=57&t=13393 and let me know for what other changes I should create pull requests, thanks.

Please forgive my mistakes, I am still figuring out how GitHub works.


Chameleonhider:
@yvt if You could please review and approve/deny these pull requests, I could commit more of them - don't want to make too many commits at once.

Issue #383, opened 2016-02-10
Cannot compile - scoreboard player palette -> ERR LNK2019


Chameleonhider:
Compiling on Win7 64 VS2013 Express I get this error:

`Error LNK2019: unresolved external symbol "int (* palette)[3]" (?palette@@3PAY02HA) referenced in function "private: void __thiscall spades::client::ScoreboardView::DrawPlayers(int,float,float,float,float)" (?DrawPlayers@ScoreboardView@client@spades@@AAEXHMMMM@Z)    C:\Projects\openspades\OpenSpades.msvc\Sources\ScoreboardView.obj   OpenSpades`

Error can be solved by removing "extern" from line ~260 in ScoreboardView.cpp, but then players won't be colored in scoreboard.
(feature introduced in https://github.com/yvt/openspades/pull/340 )

VelocityRa:
This is fixed with my PR above.

Issue #385, opened 2016-02-12
Support for Microsoft Visual Studio 2015


VelocityRa:
It seems peculiar that this project still uses VS 2013.
Pretty much everyone uses 2015, I don't want a 6.5GB download just to be able to build/contribute to the project.
How can I be the first person suggesting this?

Chameleonhider:
It is compatible, but you need libraries compiled with VS 2015 version, 2013 version libraries won't work with VS 2015. I did try compiling and it was fine except issues with '13 libraries being incompatible with VS'15.

VelocityRa:
Yeah, you do need to compile those for yourself apparently, I couldn't find any precompiled .lib files (IIRC it's SDLmain and libcurl that have the problem). I had some problems compiling the former as well. Also in order for it work I had to change a setting in the compiler's settings. 
Maybe I'll upload the libraries needed for VS 2015 and update the readme with instruction for it as well, could save other people using 2015 some time.

VelocityRa:
I can't compile libcurl.lib for the life of me, I'm not gonna waste more time on this.

PR #386, opened 2016-02-17
Experimental compatibility with VS 2015

VelocityRa:
Experimentally solves #385.

### DIFF
--- a/Sources/Gui/Main.cpp
+++ b/Sources/Gui/Main.cpp
@@ -57,13 +57,23 @@
 #endif
 
 
+#ifdef _MSC_VER>=1900	// Visual Studio 2015 or higher
+extern "C" { FILE __iob_func[3] = { *stdin,*stdout,*stderr }; }
+
+static const unsigned char splashImage[] = {
+	"SplashImage.inc"
+};
+static const unsigned char Icon[] = {
+	"Icon.inc"
+};
+#else
 static const unsigned char splashImage[] = {
 	#include "SplashImage.inc"
 };
-
 static const unsigned char Icon[] = {
 	#include "Icon.inc"
 };
+#endif
 
 
 SPADES_SETTING(cl_showStartupWindow, "1");
--- a/Sources/Gui/Main.cpp
+++ b/Sources/Gui/Main.cpp
@@ -59,21 +59,14 @@
 
 #ifdef _MSC_VER>=1900	// Visual Studio 2015 or higher
 extern "C" { FILE __iob_func[3] = { *stdin,*stdout,*stderr }; }
+#endif
 
-static const unsigned char splashImage[] = {
-	"SplashImage.inc"
-};
-static const unsigned char Icon[] = {
-	"Icon.inc"
-};
-#else
 static const unsigned char splashImage[] = {
 	#include "SplashImage.inc"
 };
 static const unsigned char Icon[] = {
 	#include "Icon.inc"
 };
-#endif
 
 
 SPADES_SETTING(cl_showStartupWindow, "1");
--- a/README.md
+++ b/README.md
@@ -94,18 +94,18 @@ GCC 4.8 / Clang 3.2 or later is recommended because OpenSpades relies on C++11 f
 
 ### On Windows (with visual studio)
 
-1. Get CMake, Visual Studio 2013 (Express), and the OpenSpades source.
+1. Get CMake, Visual Studio 2013 (Express) or Visual Studio 2015 (not very supported), and the OpenSpades source.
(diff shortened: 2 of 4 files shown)


Sources/Gui/Main.cpp
```diff
@@ -57,13 +57,23 @@
 #endif
 
 
+#ifdef _MSC_VER>=1900	// Visual Studio 2015 or higher
+extern "C" { FILE __iob_func[3] = { *stdin,*stdout,*stderr }; }
```
yvt: I don't think this works, because [`__iob_func` should be defined as a function](https://groups.google.com/forum/#!topic/dislin-users/kagVl9LVFJA) returning the array of `FILE`s.

VelocityRa: It did work for me (without any warnings or anything).
It acts as an array of pointers to stdin, stdout and stderr. `FILE` is need for some reason, instead of just a pointer, but it works.
Apparently there's still some problems with resource loading (particularly with Icon.inc and SplashImage.inc because of the way they're loaded at compile time), but this PR is a start to VS2015 compatibility, it takes care of the most important problems.

VelocityRa:
Seems like I forgot to include the link for the compiled libraries in the readme file, here it is:
https://www.dropbox.com/s/pfvk8tns1qgkxpr/OpenSpades-Externals-Windows-VS2015.rar?dl=1
Should I delete this PR and make another one?

VelocityRa:
@yvt So, should I change something? You don't like the solution? 
Also, forgot to mention, this also fixes #383.

Sources/Gui/Main.cpp
yvt: @VelocityRa I'm wondering what's wrong with the way the resources are loaded. Does it cause compile to fail? If so, can you post the compiler output?

VelocityRa: Am I doing something wrong? Maybe I need to recompile libpng as well for VS2015?
Here's the error: http://puu.sh/noxbo/85f8401b92.png

yvt: You are forgetting to put `libpng16-16.dll` (which can be found in [the OpenSpades binary release](https://github.com/yvt/openspades/releases/tag/v0.0.12)) into the same directory as the executable file.

VelocityRa: Yes, sorry, I cloned the repo again and forgot to include the dlls. [This](http://puu.sh/nozm9/640ad631ac.png) is the actual error.
But it turns out if I include the 2 images at compile time (with `#include` like it was) I do get [errors](http://puu.sh/nozA7/77a5d6302b.png) inside VS but the project actually builds... When I run it I get [this](http://puu.sh/nozBO/9127321d7e.png) error (SystemMessages.log is empty) Why could it fail to build the scripts?.

I never imagined it would be *this* difficult...

yvt: > But it turns out if I include the 2 images at compile time (with #include like it was) I do get errors inside VS but the project actually builds... 

Intellisense's problem. Let's ignore it for now.

>  When I run it I get this error (SystemMessages.log is empty) Why could it fail to build the scripts?.

I don't know. I remember that someone has reported a similar issue, but couldn't solve it because of the lack of information. Setting a breakpoint [here](https://github.com/yvt/openspades/blob/master/Sources/ScriptBindings/ScriptManager.cpp#L45) and inspecting the error message might help.

VelocityRa: I checked the filesystem (the directories it was loading stuff from, and of them was the directory where the .exe was, was, so I copied the Resources folder from the release (the .pak's) *and* the source to that dir, and it worked! The game run perfectly fine. For some reason it didn't look for the folder `OpenSpades.msvc/os.Resources` which the REAME instructs to put the .pak's in. Did I use Cmake wrong? I just did `cmake .`
Thanks for the help man, awesome project you got there.

yvt: Thank you for taking your time for this pull request.

> Edit: I suppose I should remove the changes to the images' loading now, seeing as it works that way, right?
Yeah, please revert those changes and push again.

noway421:
>For some reason it didn't look for the folder `OpenSpades.msvc/os.Resources` which the README instructs to put the .pak's in

Out of all places openspades looks for the resource files, this is an unlikable one. Probably README describes outdated behaviour. `OpenSpades.msvc/Resources` would sound like a more likely place where it will look.

VelocityRa:
Yeah, we'd either have to dig it up in the source code, or put a breakpoint like I did at the time.
Second option seems easier, but I can't get an dev environment set up (again), so I can't do it.

PR #399, opened 2016-06-17
Updated spectator behavior.

Rootmars:
When spectating, primary and secondary fire (default left and right click)
now go to the next and previous player, respectively. I did this so that the
spectator doesn't have to cycle through all the players again when he/she
"overshoots" by spamming click too much.

The button to cycle through modes is now jump (default spacebar). it goes
from "free", to "follow", to "first person". If the player is not on the spectator
team, then skips through free automatically.

Lastly, I moved the "Following {playername}" text to the top of the screen.
I also added the player's ID#. This is useful if you're spectating a possible
cheater and you need to report its ID#.

### DIFF
--- a/Sources/Client/Client.h
+++ b/Sources/Client/Client.h
@@ -138,7 +138,7 @@ namespace spades {
 			KeypadInput keypadInput;
 			Player::ToolType lastTool;
 			bool hasLastTool;
-			bool FirstPersonSpectate;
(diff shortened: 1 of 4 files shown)


Rootmars:
bug found.

NotAFile:
This is great! did you ever get around to fixing the issues with this?

PR #411, opened 2016-07-20
Fixed memory leak

myrage2000:


myrage2000:
What does that travis error mean ? It's working fine on my Visual Studio.
I'm trying to fix remaining memory leaks.

yvt:
Looks like Travis' [virtualization environment](https://docs.travis-ci.com/user/ci-environment/) has changed for some reason.

Sources/Draw/GLRenderer.cpp
```diff
@@ -256,6 +256,9 @@ namespace spades {
 			imageManager = NULL;
 			delete fbManager;
 			fbManager = NULL;
+			
+			delete device;
```
yvt: Thank you for noticing! Actually, however, this is not necessary because `device` has a type of `Handle<IGLDevice>` which is a smart pointer, and therefore it's released automatically when `GLRenderer` is destroyed. (or maybe I'm missing something...?)

myrage2000:
Yes you're right, device is leaking 0 bytes of memory, but should you call delete anyway ? I don't even know xD

myrage2000:
Oh I see I'm wrong I didn't know it was a smart pointer. My bad.

PR #413, opened 2016-08-03
Add mumble support

L0L022:
Works on linux and not tested on windows

### DIFF
--- a/Sources/Client/MumbleLink.h
+++ b/Sources/Client/MumbleLink.h
@@ -5,39 +5,41 @@
 #include <windef.h>
 #endif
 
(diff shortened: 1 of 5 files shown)


Sources/Client/MumbleLink.h
```diff
@@ -0,0 +1,41 @@
+#include "Player.h"
```
theunknownxy: You should put a header guard at the beginning of the header:
```cpp
#pragma once

...
```
And wrapping the classes into the `spades` namespace for consistency is probably a good idea.

yvt:
Cool! Thanks:)

PR #435, opened 2016-11-12
Update pt_br translation (Player name/Water Shader were added)

feikname:
Update pt_br translation (Player name/Water Shader were added to the openspades.pot file)

Translations added:
 * Player name selection upon first launch inside CreateProfileScreen.as
 * Water Shader setting inside StartupScreen.as


yvt:
Thanks @feikname!

PR #442, opened 2016-11-17
Rename 'texture' GLSL occurences for max compatibility

VelocityRa:
They were all just renamed to `texture_`.

### DIFF
--- a/Resources/Shaders/BasicImage.fs
+++ b/Resources/Shaders/BasicImage.fs
@@ -19,13 +19,13 @@
  */
 
 
-uniform sampler2D texture;
+uniform sampler2D texture_;
 
 varying vec4 color;
 varying vec2 texCoord;
 
 void main() {
-	vec4 col = texture2D(texture, texCoord);
+	vec4 col = texture2D(texture_, texCoord);
 	col.xyz *= col.w; // alpha premultiply
 	col *= color;
 	gl_FragColor = col;
--- a/Resources/Shaders/BasicImage.fs
+++ b/Resources/Shaders/BasicImage.fs
@@ -19,13 +19,13 @@
  */
 
 
-uniform sampler2D texture_;
+uniform sampler2D mainTexture;
 
 varying vec4 color;
 varying vec2 texCoord;
 
 void main() {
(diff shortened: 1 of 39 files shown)


VelocityRa:
Waiting for @yvt's OK to merge this.

yvt:
Suggestion: how about renaming them to something like `mainTexture`?

Also, I think all references (e.g., [`GLColorCorrectionFilter.cpp`:L49](https://github.com/yvt/openspades/blob/master/Sources/Draw/GLColorCorrectionFilter.cpp#L49)) to the uniform name should be renamed accordingly too. (A simple regex probably should do it, but I'm not certain)

VelocityRa:
`mainTexture` sounds good and I'll fix all references too tomorrow, my bad.
What's the easiest way to get a dev env running? VS2013? I still haven't gotten VS2015 to work...

yvt:
I just pushed [a commit](f1f88d3e183e2ce022aa4330fd425e7873190d73) that makes it possible to build successfully with VS2015. Follow the instructions in [`README.md`](https://github.com/yvt/openspades#on-windows-with-visual-studio) precisely and everything should be fine. 

It might be better to set `OPENSPADES_RESDIR` to point to the `(src root)/Resources` directory during the development, because `mkpak.sh` doesn't work on Windows yet. You also need to download the non-free pak manually.

VelocityRa:
That should do it.
Cool, I'll try building again!

VelocityRa:
Getting about 270 unresolved externals from SDL2 and GLEW... Which is weird because the .libs are there and configured properly.

yvt:
Have you tried deleting the CMake cache (File → Delete Cache) and running CMake again? If you have previously run it without lib files, it won't search them again and won't report any errors until you delete the cache or relevant cache entries.

VelocityRa:
Good call, I'll try that.
Is email a good way to contact you? Since all this is irrelevant to the issue.

yvt:
Yes.  My email address is on my profile page.

VelocityRa:
Shit I didn't notice you rebased the commits, sorry.

PR #455, opened 2016-11-20
Fix default values on GCC/stdc++/glibc.

theunknownxy:
c_str() returns an null pointer for empty strings on stdc++(GCC/clang on
linux) and glibc causes a segfault if an null pointer is passed to atof/atoi.

### DIFF
--- a/Sources/Core/Settings.cpp
+++ b/Sources/Core/Settings.cpp
@@ -307,8 +307,16 @@ namespace spades {
                 item->descriptor = descriptor;
                 const std::string &defaultValue = descriptor->defaultValue;
                 if (item->defaults) {
-                    item->value = static_cast<float>(atof(defaultValue.c_str()));
-                    item->intValue = atoi(defaultValue.c_str());
+                    // c_str() returns an null pointer for empty strings
+                    // on stdc++(GCC/clang on linux) and glibc causes a segfault if
+                    // an null pointer is passed.
+                    if (defaultValue.c_str() != nullptr) {
+                        item->value = static_cast<float>(atof(defaultValue.c_str()));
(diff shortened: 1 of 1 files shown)


yvt:
That's strange. How could `c_str()` return `nullptr` for empty strings in one place while not in others?

My guess is that, the static variable initialization order is causing this problem. 

[Settings.h](https://github.com/yvt/openspades/blob/master/Sources/Core/Settings.h#L151):
```c++
// Define SettingItemDescriptor with external linkage so duplicates are
// detected as linker errors.
#define DEFINE_SPADES_SETTING(name, ...) \
    spades::SettingItemDescriptor name ## _desc { __VA_ARGS__ }; \
    static spades::Settings::ItemHandle name ( # name , & name ## _desc )
    
#define SPADES_SETTING(name) \
    extern spades::SettingItemDescriptor name ## _desc; \
    static spades::Settings::ItemHandle name ( # name , & name ## _desc )
```

`DEFINE_SPADES_SETTING` and `SPADES_SETTING` are done in separate compilation units, which means the initialization of `SettingItemDescriptor` and `ItemHandle` is [done in an undefined order](http://stackoverflow.com/questions/211237/static-variables-initialisation-order). A possible outcome of this is that, `ItemHandle` gets initialized with a not yet initialized `SettingItemDescriptor`, which could lead to a strange behavior.

Instead of null-checking the return value of `c_str()`, I suggest rewriting the above lines of code to:

```c++
#define SPADES_SETTING(name) \
    static spades::Settings::ItemHandle name ( # name , nullptr )
```

theunknownxy:
@yvt Sorry I was pretty busy in the last few days and didn't have enough time looking at your proposed change. Thanks for doing a fixed pull request(#464).

PR #462, opened 2016-11-23
Move [Build Status] next to repo name

skmedix:
Cause Build Status in current place just looks weird


VelocityRa:
Not sure about this, it looks like it's part of the title.

Issue #436, opened 2016-11-14
Favorites List  


Danogot516:
We should have a "Favourites List" in which we can bookmark our favourite servers so we can easily find them when we want to play in them.

noway421:
It's not like there are many servers nowadays though.

pielover88888:
Let's do what Iceball did - not let the community shrinking impact progress

feikname:
Solved in #465.

PR #465, opened 2016-11-30
Favorite Servers

theunknownxy:
This is a implementation of #436.

TODO
- [x] How to mark favorite servers?
- [x] How to render favorite servers?

### DIFF
--- a/Sources/Gui/MainScreenHelper.h
+++ b/Sources/Gui/MainScreenHelper.h
@@ -62,7 +62,7 @@ namespace spades {
 			int GetNumPlayers() { return numPlayers; }
 			int GetMaxPlayers() { return maxPlayers; }
 			bool IsFavorite() { return favorite; }
-			bool SetFavorite(bool favorite) { this->favorite = favorite; }
+			void SetFavorite(bool favorite) { this->favorite = favorite; }
 		};
 
 		struct MainScreenServerList {
--- a/Sources/Gui/MainScreenHelper.h
+++ b/Sources/Gui/MainScreenHelper.h
@@ -20,6 +20,7 @@
 
 #pragma once
 
+#include <unordered_set>
 #include <vector>
 
 #include <Core/Mutex.h>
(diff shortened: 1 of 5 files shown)


theunknownxy:
Tinting the background of favorite servers looks pretty decent.
Highlighting
Any opinions for the color?

VelocityRa:
Took me a good 10 seconds to notice it, make it more prominent!

feikname:
How about keeping the blue background and changing text color to yellow/gold? Yellow and blue go good together!

 The same color that @VelocityRa used in the Spectator list sounds good to me.
image

theunknownxy:
Here you go:
Yellow!

noway421:
Just say when it's ready as it is now marked as [WIP]. Looks really good.

theunknownxy:
If nobody has any suggestions it's ready. 
Also you can't change labels of your issues if you aren't a collaborator. What about removing the "work-in-progress" label and instead use the convention that prefixing a PR with "WIP: " means it's work in progress?

feikname:
Me and @VelocityRa log in almost everyday to ensure tagging is correct. 

But I have to agree, it's a pain in the ass for people wanting to submit a pull request that they can't change their own tags. The suggestion of modifying the title (or just the contents, really, it's rare to have more than 4 issues pull requests open and it's fast to open and see their status) sounds good to me.

@VelocityRa should we remove the `PR: work-in-progress` label?

Sources/Gui/MainScreenHelper.h
```diff
@@ -59,6 +61,8 @@ namespace spades {
 			int GetPing() { return ping; }
 			int GetNumPlayers() { return numPlayers; }
 			int GetMaxPlayers() { return maxPlayers; }
+			bool IsFavorite() { return favorite; }
+			bool SetFavorite(bool favorite) { this->favorite = favorite; }
```
yvt: `return` is missing in this non-void function. (specified by C++ spec as undefined behavior)

theunknownxy: Oops it's not only undefined behavior, it also makes no sense.

VelocityRa:
@feikname @theunknownxy **WIP:** sounds good to me. Let's ditch the "PR:work-in-progress" label.

feikname:
@VelocityRa done.

yvt:
Thank you for fix. Looks very good.

Issue #412, opened 2016-07-22
Allow spectators to be displayed on the scoreboard


100best:
I have noticed that openspades right now does not have something to show spectators like the classic game did.

sanchez-k:
this add-on will be cool

Tapmemer:
This is easily sorted out by showing all ids and name of team 2(specs used to be team 255)

100best:
What I mean is the classic game had a place to show all spectators.

Tapmemer:
Exactly what i said

feikname:
May I suggest to rename the issue to "Allow spectators to be displayed on the scoreboard"?

VelocityRa:
I'm gonna have a go at this, where do we want spectators placed?

Here's the scoreboard as of now:
![image](https://cloud.githubusercontent.com/assets/6632271/20725181/a8b84802-b679-11e6-8a13-feb208ac513b.png)

I was thinking about listing them on the bottom horizontally? I can make the background taller if need be.

feikname:
just out of curiosity. What happens if there are more players than OpenSpades can display? is there a scrollbar?

VelocityRa:
That height accomodates the max number of players that can be in a team, so no scrollbar necessary.

VelocityRa:
It displays them like so:

![image](https://cloud.githubusercontent.com/assets/6632271/20726362/eb1e00e8-b67d-11e6-86e9-62c5e78dd3f5.png)

feikname:
Ah yeah. I forgot that. Thanks!

feikname:
I like your idea of displaying horizontally. Reminds me of [Team fortress 2 Scoreboard](https://wiki.teamfortress.com/wiki/Scoreboard/pt-br)

VelocityRa:
## Current progress
(they display green team's players for debugging)

![image](https://cloud.githubusercontent.com/assets/6632271/20729692/cabe10d2-b68b-11e6-9152-426635285a54.png)
(using the server's color)


![image](https://cloud.githubusercontent.com/assets/6632271/20729814/52643f5c-b68c-11e6-8e18-7e1c6538842a.png)
(using yellow)

As you can see they align to the center, it's more visually appealing IMO.
Thoughts?
@yvt

VelocityRa:
Here's another one, I improved the center aligning:

[...]

PR #466, opened 2016-11-30
Display Spectators in Scoreboard

VelocityRa:
For discussion see the relevant Issue: #412.

Current look:

### DIFF
--- a/Sources/Client/ScoreboardView.cpp
+++ b/Sources/Client/ScoreboardView.cpp
@@ -41,6 +41,12 @@ SPADES_SETTING(cg_Minimap_Player_Color);
 
 namespace spades {
 	namespace client {
+
+		static const Vector4 white = { 1, 1, 1, 1 };
+		static const Vector4 spectatorIdColor = { 210.f / 255, 210.f / 255, 210.f / 255, 1 }; // Grey
+		static const Vector4 spectatorTextColor = { 220.f / 255, 220.f / 255,0,1 };	          // Goldish yellow
+		static const auto spectatorTeamId = 1;		// Spectators have a team id of 255
+
 		ScoreboardView::ScoreboardView(Client *client):
 		client(client), renderer(client->GetRenderer()) {
 			SPADES_MARK_FUNCTION();
@@ -126,6 +132,7 @@ namespace spades {
 			float contentsLeft = scrWidth * .5f - 400.f;
 			float contentsRight = scrWidth * .5f + 400.f;
 			float playersHeight = 300.f;
+			float spectatorsHeight = 78.f;
 			float playersTop = teamBarTop + teamBarHeight;
 			float playersBottom = playersTop + playersHeight;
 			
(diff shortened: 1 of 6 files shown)


Sources/Client/ScoreboardView.cpp
```diff
@@ -297,5 +313,73 @@ namespace spades {
...
+
+				numSpectators++;
+
+				// Measure total width in pixels so that we can center align all the spectators
+				sprintf(buf, "#%d", ent.id);
+				totalPixelWidth += font->Measure(buf).x + font->Measure(ent.name).x + xPixelSpectatorOffset;
+			}
+			switch (numSpectators) {
+				case 0: return;
+				case 1:	strcpy(buf, "Spectator:");
```
theunknownxy: I think you can use [GetTextPlural](https://github.com/yvt/openspades/blob/master/Sources/Core/Strings.h#L267) to make it more localization friendly.

VelocityRa: Good catch, I'll fix that, thanks.

feikname: Was about to say that

VelocityRa: "Spectator" doesn't belong in the localization lists though, it's not translated in any language.

theunknownxy: Not yet?

VelocityRa: Ah, this function literally returns the plural depending on the language, right, nvm.

feikname: Could you please tell how you did it? I was wondered :)

My guess is update-pot.sh

VelocityRa:
There we go fixed.
Having second thoughts about that colon now... It looks fine but it's not consistent with the the teams' texts.

feikname:
I think it looks nice with the colon.

feikname:
@VelocityRa there is something wrong with your implementation of plural forms, it always stays on the singular, tested it right now.
image

VelocityRa:
@feikname Have you specified the plural forms [correctly](https://github.com/yvt/openspades/blob/master/Resources/Locales/de/openspades.po#L1573)? It should work.

feikname:
I tried to simplify the line to `strcpy(buf, _TrN("Client", "Spectator", "Spectators", numSpectators).c_str());`, updated the `.pot` file and double-check my translation and is still not working. 

The most weird part is that without any`core_locale` it works fine.

VelocityRa:
Based on other usages, I *think* I'm using this macro correctly.
Gonna have to ask @yvt's help on [this](https://github.com/yvt/openspades/pull/466/commits/ac35fcea0811396aae074b7d6e7bc86fe3d7c4ce).

feikname:
I will double check myself too. I am almost 100% sure you did the right implementation, but I can't get it to work. Perhaps it's some dark auto-casting happening with NumSpectators

VelocityRa:
Test it with `1` and `2` instead of `numSpectators`. Does it work correctly then?

feikname:
I can only try a few hours from now, I will test as soon as I arrive back home

feikname:
Ok, I will try now.

feikname:
@VelocityRa status update:

Ok, I tested and uploaded my changes to [my fork](https://github.com/feikname/openspades) of your fork.

I made sure to double-check the translations and also named them them as `Singular{1}` and `Plural{1}`, so I am almost sure there's no way I wronged them. In case you wanna see for yourself, this is [the commit] where I created them](https://github.com/feikname/openspades/commit/ff81607c3d7f66ff13b566b4f3b1fc171fbcd655). 

The `.pot` was generated from `update-pot.sh` (xgettext) and the `pt_br/openspades.po` was edited from *Poedit*, so I am exluding any syntax errors too. `update-pot.sh` was run from *Bash on Windows* (lxss).

I also used "2" as `numSpectators`, so it's not a issue with `numSpectators` value.

**When my `core_locale` is set to "pt_br", this is the result:**
image

**However, with `core_locale` set to empty (nothing), it works correctly 😒**
image

----

This is very weird, maybe a new issue about this should be opened? If you can confirm on your system too, I'd be very grateful.

VelocityRa:
>This is very weird, maybe a new issue about this should be opened? If you can confirm on your system too, I'd be very grateful.

Yeah it's present on my system too.

VelocityRa:
@yvt Rebased your commits to test the locale thing and it works now!
Should I merge this?

PR #472, opened 2016-12-03
Fix compilation on newer libstdc++.

theunknownxy:
Newer versions only declare isnan in the std namespace when the cmath header is used.


noway421:
I remember there was similar problem with some builds of MSVC, so this PR brings it closer in terms of support across compilers.

PR #474, opened 2016-12-04
Move building instructions outside README.md

feikname:
This pull moves the building instructions to outside the README.md file. 

The intention is to tidy up the README.md file, making it more clean.

A preview of the results can be seem [here](https://github.com/feikname/openspades/tree/move-building-instructions)


VelocityRa:
Not sure about [this](https://github.com/feikname/openspades/tree/move-building-instructions) tbh.
 
The README should be as helpful as possible without extra clicks IMO, and building instructions is one of the most important things it must contain. We _can_ try and make the building guide more succinct, however. The Windows guide for example can be made smaller by replacing it with the wiki's one (and provide a link to the wiki for the longer one.

feikname:
I think you're right. It seems too empty.

 Not sure if the linux one can be made any smaller tho.

noway421:
Depends on your target. The instruction is mainly for ubuntu (12.02 or was it), for a more fresh distro you wouldn't need to download and compile sdl2 and sdl2_image on your own.

all the "**note**:" are not really needed anymore

there has been some changes to resource loading and paks too, that might open up a way to describe their install into more clear way, but not really sure about that.

PR #485, opened 2016-12-10
Added the Portuguese language and fixed typos on POT and pt_BR

DarkDracoon:
I'm requesting this pull in order of adding support to **Portuguese** language and fixing typos found on **POT** - base language (which caused to change on the other languages) and **pt_BR**.
Additionally, on the JSON that identifies the languages, I've changed "_Portuguese_" to "_Portuguese (Brazil)_" and created the pt_PT.json to identify the _Portuguese (Portugal)_ language.

_PS: There might be things that can be slightly equal between Portuguese (Portugal) and Portuguese (Brazil), since both languages are based from the same language, only the regional changes._

### DIFF
--- a/Resources/Locales/pt_br/openspades.po
+++ b/Resources/Locales/pt_br/openspades.po
@@ -544,7 +544,7 @@ msgstr "Ver."
 #: ../../Scripts/Gui/MainScreen.as:515 ../../Scripts/Gui/MainScreen.as:529
 msgctxt "MainScreen"
 msgid "Loc."
-msgstr "Ling."
+msgstr "Loc."
 
 #: ../../Scripts/Gui/MainScreen.as:767 ../../Scripts/Gui/MainScreen.as:787
 msgctxt "MainScreen"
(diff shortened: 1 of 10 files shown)


Resources/Locales/pt_br/openspades.po
```diff
@@ -544,7 +544,7 @@ msgstr "Ver."
 #: ../../Scripts/Gui/MainScreen.as:515 ../../Scripts/Gui/MainScreen.as:529
 msgctxt "MainScreen"
 msgid "Loc."
-msgstr "Loc."
+msgstr "Ling."
```
feikname: @DarkDracoon "Loc." represents the physical location of the server, not the language.

Also, I am the translator of the pt_br language, thanks for the fixes. Submit a pull request whenever you find something that can be improved :)

DarkDracoon:
Sorry for that. Just corrected the source text and let the **Loc.** be as it was.
I just started looking for code usages after I discovered the code browser on _Poedit_, that's why I didn't see where was **Loc.** being used.

PR #502, opened 2016-12-25
Improved Korean translation

kmsiapps:
* Removed some blanks on the translation to fit original translation.
* Improved some translation with Antialias


yvt:
Thanks!

PR #504, opened 2016-12-25
Improve update-pot.sh and fix #503

### LINKED ISSUE #503: update-pot.sh doesn't include changes in non .cpp files
Proof:
 1. Change any non .cpp file (e.g. `Resources/Scripts/Gui/MainScreen.as`)
 2. Run update-pot.sh
 3. See no changes happen

This happens on bash.

The culprit is the `!#/bin/sh` [shebang](https://en.wikipedia.org/wiki/Shebang_(Unix)), because it forces bash to run on portability mode, which does not support the `+=` syntax for strings concatenation.

feikname:
Changing the [Shebang](https://en.wikipedia.org/wiki/Shebang_(Unix)) to `#!/usr/bin/env bash` also works, as this way bash leaves the compatibility mode and supports the `+=` syntax. However, I think the file readability and the portability are better this way. This pull fixes #503 for me.

I don't know if it works on macOS, but it's very likely.

Please note that the outdated Antialias translations still need to be fixed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -69,5 +69,6 @@ openspades.mk/
 Sources/Gui/DetailConfigWindow.txt
 Sources/Gui/MainWindow.txt
 
-# Compiled .mo files (translation)
+# Translation files
 *.mo
+.translate.this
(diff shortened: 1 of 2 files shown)


noway421:
Good job on that fix. Although I'd adore `/bin/sh`, in 2017 `bash` would be well enough to rely to in terms of interoperability.

PR #505, opened 2016-12-26
Make translations consistent

feikname:
This make all translations consistent. Changes:
 * Headers are now the same on all translation, with copyright updated to 2016
 * All translations are now updated to match the .pot file, in case a new translator without the knowledge of how to update from catalog tries to translate (it's not specified in the wiki yet)
 * Fixed the issues with oudated Antialias translations because of #503

The changes were tested on GNU/Linux *ElementaryOS 0.4*, everything seems to be fine.


Resources/Locales/pt_br/openspades.po
```diff
@@ -1105,24 +1077,23 @@ msgid ""
...
 msgstr ""
-"Antialias: Habilita uma técnica para melhorar a aparência de bordas de alto "
+"Antialias:Habilita uma técnica para melhorar a aparência de bordas de alto "
 "contraste.\n"
 "\n"
-"MSAA: Executa anti-aliasing gerando uma imagem intermediária de alta "
-"resolução. Têm a melhor aparência, mas pode não funcionar bem com certas "
-"configurações.\n"
+"MSAA: Executa anti-aliasing gerando uma imagem intermediária de alta \"\n"
+"\"resolução. Têm a melhor aparência, mas pode não funcionar bem com certas "
```
yvt: `\"` looks unnecessary here...

feikname: That was not supposed to happen.

PR #506, opened 2016-12-26
Russian translation

ArionWT:
Actual version of Russian translation


ArionWT:
Updated. Shorted some phrases.
[openspades.zip](https://github.com/yvt/openspades/files/673773/openspades.zip)
Russian translation looks better on [Roboto Condensed Regular](https://fonts.google.com/specimen/Roboto+Condensed), than NotoSans.

And one more question, can you make, that custom font isn't applied only to new symbols, but also to the Latin symbols? To avoid apart text look.

feikname:
@ArionWT I added a commit for your changes.

yvt:
Thanks!

PR #510, opened 2016-12-31
Added depth-only prepass.

roboman2444:
OVM does the prepass
Not sure if it actually helps performance, since OS renders the world front->back for the most part.

### DIFF
--- a/Resources/Shaders/BasicBlockDepthOnly.fs
+++ b/Resources/Shaders/BasicBlockDepthOnly.fs
@@ -1,0 +1,26 @@
+/*
+ Copyright (c) 2013 yvt
+ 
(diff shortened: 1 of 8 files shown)


yvt:
Good work! 👍 

I'm uncertain about the performance impact of this addition because the last time (2 yrs ago) I was tweaking the map renderer it was mostly draw call bound. Maybe I'm going to add an option to toggle depth pre-pass later. In any case, the depth pre-pass will be crucial to implement SSAO or whatever screen-space global illumination algorithm.

roboman2444:
Have you thought about adding some more advanced OpenGL features to speed up and reduce the draw calls? Mainly VAOs and instancing?

PR #518, opened 2017-01-02
Update pt_br openspades.po

feikname:


VelocityRa:
Why do you update this so often?

feikname:
@VelocityRa I like to keep the translations up to date, sorry for that.

I will submit pull request only when OpenSpades is about to release from now on.

feikname:
Also, I don't update that often. I just update when new strings are added, which happened to be in very close intervals this time, but usually isn't.

VelocityRa:
Sounds like a plan.

Sorry if I sounded abrupt earlier, but it's not just me that you have to consider, there's 57 "watchers" in this repo.

Issue #507, opened 2016-12-26
Notofont doesn't seem to be applied on "Spectator" string at scoreboard


kmsiapps:
![shot0000](https://cloud.githubusercontent.com/assets/23615360/21482687/ac110e20-cbba-11e6-965c-793ffd5fb685.jpg)
Yellow text saying "관전자"(means Spectator in Korean) is still with default unicode font, even though I downloaded Notofont. Maybe it would be a bug ;)

VelocityRa:
That's the case because I use the same font that BLUE/GREEN uses above.
It's intended behaviour, so that its font is consistent one used for the teams' titles.

@yvt Do you think that *if* `client->fontManager->GetGuiFont()` is not the default font, we should use that, else fall back to the present one (`client->fontManager->GetSquareDesignFont()`)?

VelocityRa:
@yvt Feedback on this? ^

VelocityRa:
@yvt Yeah that sounds good. I'll get on it.
Where should that check happen though? Can the font be chanced at any point in the game (so I'll need to check every frame), or can I check it before that?

yvt:
@VelocityRa Checking it every frame would be simpler (and is unlikely to break because of future changes), but it's also okay to check it in advance, for example, in the constructor.

VelocityRa:
Apparently I need to compile a font to use it? @kmsiapps can you send me your font so that I can test my fix?

kmsiapps:
@VelocityRa I used default notofont.pak, which is downloadable in Openspades distribute page.
Thanks for the fix :)

VelocityRa:
Without the fix:
![image](https://cloud.githubusercontent.com/assets/6632271/21619474/8326dfbe-d1f8-11e6-942a-0fe707ef4ba1.png)

With fix:
![image](https://cloud.githubusercontent.com/assets/6632271/21619487/8d1b888a-d1f8-11e6-9b72-8fa871078661.png)

I'll push it later.

VelocityRa:
Without fix:
![image](https://cloud.githubusercontent.com/assets/6632271/21619573/f4ef4d02-d1f8-11e6-83cf-52baafcdd472.png)

With fix:
![image](https://cloud.githubusercontent.com/assets/6632271/21619575/f72bedb4-d1f8-11e6-96ff-15acac97d83f.png)

I'll push it later.

Edit: This is from `GetLargeFont`.
[...]

PR #527, opened 2017-01-04
Use GUI font if spectator string contains special chars.

VelocityRa:
For #507.
Also made some `ScoreboardView` methods `const`.

### DIFF
--- a/Sources/Client/ScoreboardView.cpp
+++ b/Sources/Client/ScoreboardView.cpp
@@ -49,17 +49,32 @@ namespace spades {
 		static const Vector4 spectatorTextColor = {220.f / 255, 220.f / 255, 0,
 		                                           1}; // Goldish yellow
 		static const auto spectatorTeamId = 255;       // Spectators have a team id of 255
+		
 
 		ScoreboardView::ScoreboardView(Client *client)
 		    : client(client), renderer(client->GetRenderer()) {
 			SPADES_MARK_FUNCTION();
-			ctf = NULL;
-			tc = NULL;
+			world = nullptr;
+			tc = nullptr;
+			ctf = nullptr;
+			image = nullptr;
+
+			// Use GUI font if spectator string has special chars
+			auto spectatorString = _TrN("Client", "Spectator{1}", "Spectators{1}", "", "");
+			auto has_special_char = 
+				std::find_if(spectatorString.begin(), spectatorString.end(),
+				[](char ch) {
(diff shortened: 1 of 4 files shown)


Sources/Client/ScoreboardView.cpp
```diff
@@ -49,17 +49,33 @@ namespace spades {
...
+			tc = nullptr;
+			ctf = nullptr;
+			image = nullptr;
+
+			// Use GUI font if spectator string has special chars
+			auto spectatorString = _TrN("Client", "Spectator{1}", "Spectators{1}", "", "");
+			auto has_special_char = 
+				std::find_if(spectatorString.begin(), spectatorString.end(),
+				[](char ch) {
+					ch = static_cast<unsigned char>(ch);
```
yvt: I don't think this cast is necessary...

VelocityRa: It is, the `char` `isalnum` should be unsinged. A runtime error occurs otherwise.

yvt: Okay, I see, `isalnum` accepts a `int` and it must be representable as `unsigned char` (or is equal to `EOF`). But casting `ch` to `unsigned char` and then reassigning the result to `ch` doesn't make `ch` unsigned. In fact, this is what I get from LLDB:

```
(lldb) bt 1
* thread #1: tid = 0x1054e35, 0x00000001003733c2 OpenSpades`isalnum(_c=-24) + 18 at ctype.h:214, queue = 'com.apple.main-thread', stop reason = step in
    frame #0: 0x00000001003733c2 OpenSpades`isalnum(_c=-24) + 18 at ctype.h:214
```

yvt:
Also, can you make the text a little bit smaller to match the original font, and move it downward? The best way to add a new font size would be adding a new member function to `FontManager` (in `Fonts.h`).

Sources/Client/ScoreboardView.cpp
VelocityRa: Still, this line does _something_ that makes it work. Without it I get this when I have a font with non-ascii chars:
image
And with it it works as intended.

VelocityRa:
Can I make a new IFont member `mediumFont`(or something similar) instead? I don't have much time for this right now.

yvt:
Okay

VelocityRa:
@yvt How's that?
image

yvt:
@VelocityRa That looks great!

VelocityRa:
Cool, I force pushed to the previous commit (and made a seperate one for the font).

PR #528, opened 2017-01-04
Update VersionInfo.cpp

feikname:
* Remove use of deprecated API
 * Add support for Windows 10+ detection

Also, indented the macros a bit.

**Note:** I didn't actually test, as I have no available Windows right now.<br> Please someone test it with Visual Studio.

### DIFF
--- a/Sources/Core/VersionInfo.cpp
+++ b/Sources/Core/VersionInfo.cpp
@@ -1,11 +1,12 @@
 #if __linux__
-#define OS_PLATFORM_LINUX
+	#define OS_PLATFORM_LINUX
 #elif TARGET_OS_MAC
-#define OS_PLATFORM_MAC
+	#define OS_PLATFORM_MAC
 #elif defined _WIN32 || defined _WIN64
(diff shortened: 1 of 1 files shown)


VelocityRa:
* You have an extra brace at the end.

* `IsWindows10OrGreater` is undefined in the Windows 8.1 SDK. I can't compile it. I do have the 10 SDK but it finds the 8.1 one header first.

You need `VersionHelpers.h` from the Windows 10 SDK for it to work, and even then it will still report false unless "the application contains a manifest that includes a compatibility section that contains the GUID that designates Windows 10", which is detailed [here](https://msdn.microsoft.com/en-us/library/windows/desktop/dn481241(v=vs.85).aspx).

If you're not willing to do that (_I'm_ surely not), we could resort to defaulting to Windows 10 if all checks fail. It will be right the vast vast majority of the time anyway.

VelocityRa:
Can you allow me push access?

feikname:
@VelocityRa thanks!

Issue #524, opened 2017-01-04
Reduce redundant GL state changes.


roboman2444:
Adding a little check like 
if(curprogram == wantedprogram) return;
can sometimes help performance.

I might submit a pull request in a day or two that does this if I have time.

Also, using a "state manager" might make it easier to program with.

output of gloverrides:
81 frames in 1006.950012 milliseconds = 80.440933 fps
State status:
	glActiveTexture redundant/total:	0/0
	glAlphaFunc redundant/total:	0/0
	glBindTexture redundant/total:	4241/50027
	glBindVertexArray redundant/total:	0/0
	glBlendFunc redundant/total:	567/891
	glColor* redundant/total:	0/0
	glCullFace redundant/total:	0/0
	glDepthFunc redundant/total:	324/810
	glDepthMask redundant/total:	162/486
	glUseProgram redundant/total:	7142/10301

PR #538, opened 2017-01-09
WIP Reduce redundant state changes.

roboman2444:
WIP, but its probably better if someone with more knowledge of the engine finished it...

Still needs texture, framebuffer, buffer state management.

Works on #524

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ GCC 4.9 / Clang 3.2 or later is recommended because OpenSpades relies on C++11 f
    *On Debian-derived distributions*: 
    ```
    sudo apt-get install pkg-config libglew-dev libcurl3-openssl-dev libsdl2-dev \
-     libsdl2-image-dev libalut-dev xdg-utils libfreetype6-dev libopus-dev libopusfile-dev
+     libsdl2-image-dev libalut-dev xdg-utils libfreetype6-dev libopus-dev libopusfile-dev cmake
    ```
    (because of a bug in some distributions, you might also
    have to install more packages by `sudo apt-get install libjpeg-dev libxinerama-dev libxft-dev`)


roboman2444:
I'm gonna need to know the max texture units in order to do no redundant texture bindings.

yvt:
The water renderer uses up to 16 texture units, which is the maximum number of texture units supported by macOS's OpenGL compatible profile driver (no matter how many units are supported by the hardware).

roboman2444:
Is there any performance or quality improvement that can be made by using more texture units if the hardware/software supports it?

yvt:
While making the level 3 water renderer (`r_water 3`) I had to drop one dynamic water texture because of 16 texture units limitation, and as a consequence the repeating pattern is more visible than that of the level 2 renderer. Also, the level 2 and 3 water renderers use a special version of a lighting shader named `CommonNoSSAO` to reduce the texture unit usage.

Except that, the limitation has never been a problem.

roboman2444:
Have you thought about possibly using texture arrays for that?
You also only use 2 components from wavetexture2 and wavetexture3. You could possibly pack them into a single 4 component texture.

yvt:
Oh, why didn't I notice that? Thanks for pointing that out!

Sources/Gui/SDLGLDevice.h
```diff
@@ -25,16 +25,54 @@
...
+			float lw_w;
+			//DepthFunc
+			spades::draw::IGLDevice::Enum df_func;
+			//UseProgram
+			spades::draw::IGLDevice::UInteger up_program;
+			//Texture stuff
+			unsigned char tex_active;
+			//change the number if you want more texture units!
+			spades::draw::IGLDevice::UInteger tex_id[SPADES_SDLGL_TEXTURE_UNITS];
+			spades::draw::IGLDevice::Enum tex_target[SPADES_SDLGL_TEXTURE_UNITS];
```
yvt: I'm no expert of OpenGL, but it seems that OpenGL doesn't work like that:

https://www.khronos.org/opengl/wiki/Texture#Texture_image_units

> Each texture image unit supports bindings to all targets. So a 2D texture and an array texture can be bound to the same image unit, or different 2D textures can be bound in two different image units without affecting each other. So which texture gets used when rendering? In GLSL, this depends on the type of sampler that uses this texture image unit.

roboman2444:
For buffers im gonna need more information on how buffers are used in AOS.
I may also suggest using VAOs. It might give a slight performance improvement, and it is required to have a VAO bound to conform to 3.3 core spec.

yvt:
> For buffers im gonna need more information on how buffers are used in AOS.

All buffer objects are merely plain old VBOs containing either vertex data or indices. There's no read backs nor feedbacks.

> I may also suggest using VAOs. It might give a slight performance improvement, and it is required to have a VAO bound to conform to 3.3 core spec.

Okay, I'll give it a try.

yvt:
It turned out that we can't use VAOs on macOS, unless we switch to the core profile and port all shaders to GLSL 1.50 or later, but doing so will force us to drop the support for pre-OpenGL 3.2 hardware.

PR #543, opened 2017-01-11
Spanish translation

brunodiazu:
i have ready a spanish translate

### DIFF
--- a/Resources/Locales/es_es.json
+++ b/Resources/Locales/es_es.json
@@ -1,0 +1,4 @@
+{
+    "descriptionEnglish": "Español (Spanish)",
+    "description": "Español (España)"
+}
--- a/Resources/Locales/es_es.json
+++ b/Resources/Locales/es_es.json
@@ -1,4 +1,4 @@
 {
-    "descriptionEnglish": "Español (Spanish)",
+    "descriptionEnglish": "Spanish",
     "description": "Español (España)"
 }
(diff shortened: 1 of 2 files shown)


Resources/Locales/es_es.json
```diff
@@ -0,0 +1,4 @@
+{
+    "descriptionEnglish": "Español (Spanish)",
```
feikname: The description here should be totally in english. 

e.g.: `"descriptionEnglish": "Spanish (from Spain)` or just *Spanish* is fine.

Resources/Locales/es_es/openspades.po
```diff
@@ -0,0 +1,1538 @@
...
+
+#: Sources/Client/Client_NetHandler.cpp:101
+msgctxt "Client"
+msgid "{0} captured an Neutral Territory"
+msgstr "{0} ha capturado un Territorio Neutral"
+
+#: Sources/Client/Client_NetHandler.cpp:127
+msgctxt "Client"
+msgid "{0} captured {1}'s intel"
+msgstr "{0} capturó la intel {1}"
```
feikname: I only know very basic spanish, but shouldn't the translation be "{0} capturó la intel de {1}"?

Then again, there is a high chance that I might be wrong, because I don't know much spanish.

Resources/Locales/es_es.json
brunodiazu: Ohh ok! Sorry, no problem! :D

Resources/Locales/es_es/openspades.po
brunodiazu: Basically is the same, anyway I will change it because they understood. Thank you :D :+1:

feikname:
Seems like everything is fine now, thanks for the contribution! 👍 

Your translation will probably be incorporated in the next version ;)

PR #548, opened 2017-01-14
While loading a map, a tack overflow is triggered either

devnexen:
in NetLog or LogMessage so here we put some limit to
the buffer. Raised on OpenBSD with gcc 4.9.4


yvt:
Thanks for your contribution!:+1:

PR #557, opened 2017-01-22
Added OPENSPADES_RESOURCES switch to install

Mailaender:
to avoid legal problems on @openSUSE.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -223,15 +223,17 @@ if(UNIX AND NOT APPLE)
 	# copyright file
 	INSTALL(FILES ${CMAKE_BINARY_DIR}/Resources/Unix/Debian/copyright DESTINATION ${OPENSPADES_INSTALL_DOC}/)
 
-	# install asset paks (including non-GPL one)
-	install(FILES
-		${CMAKE_BINARY_DIR}/Resources/pak000-Nonfree.pak
-		${CMAKE_BINARY_DIR}/Resources/pak002-Base.pak
-		${CMAKE_BINARY_DIR}/Resources/pak005-Models.pak
-		${CMAKE_BINARY_DIR}/Resources/pak010-BaseSkin.pak
-		${CMAKE_BINARY_DIR}/Resources/pak050-Locales.pak
-		${CMAKE_BINARY_DIR}/Resources/pak999-References.pak
-		DESTINATION ${OPENSPADES_INSTALL_RESOURCES})
+	if(OPENSPADES_RESOURCES)
+		# install asset paks (including non-GPL one)
+		install(FILES
+			${CMAKE_BINARY_DIR}/Resources/pak000-Nonfree.pak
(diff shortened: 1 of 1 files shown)


feikname:
I never understood what the `OPENSPADES_RESOURCES` option on CMake does.

The description says "No OpenSpades Resource", while the name clearly indicates othewise.

Mailaender:
Guess it was introduced to save some build seconds for the Travis CI system. Note: I have to change this a bit to actually suit my needs. Downloads during isolated VM builds is forbidden on OBS and I can't take the non-free dependencies. A download into $HOME on first game start step would be helpful.

damienflament:
@Mailaender : Downloading resources doesn't seems to be good for me.
I think it's an issue related to package management. And it has to be resolved by package management.
Usually, non-free data is hosted on non-free dedicated package repositories. So, you need to create a separated package which provides only non-free resources.

The perfect solution is to make free replacement resources for OpenSpades ...

Mailaender:
> So, you need to create a separated package which provides only non-free resources.

Hmm, that means submitting my package to the http://packman.links2linux.de/ and for you a lot less visibility. My idea was to get you featured in the https://software.opensuse.org/appstore but for that you will need to get rid of https://github.com/yvt/openspades-paks/blob/master/Nonfree/LICENSE.md

noway421:
Good idea to get featured!

damienflament:
@Mailaender : Can you create a FOSS package submitted to openSUSE appstore and another non-free package ?

I think that's what being done for open source game engine implementations requiring original copyrighted game files.

feikname:
Seems like the best solution is to use an external download. The only problem I see is where the download should be placed (local user resource directory?!), and the increased complexity of adding an Download GUI and a simple JSON file indicating to each Non-free Pak the OS version it belongs.

On the other hand, this might allow us to provide pak updates. But realistically it'd only be useful for translations or slowly replacing the non-free pak.

I think we should keep in mind that when distributing the packages to operating systems, unless we are getting our package into AUR, most of the time that version will stay frozen during a long period for the operating system users, so having an program that can auto-update itself is very advantageous. 

(PPAs would be an solution for apt-based systems too. I started working on one for OS a long time ago and then I forgot about it)

damienflament:
> most of the time that version will stay frozen during a long period for the operating system users

I didn't think about it. I'm not familiar with that type of distros since I'm using ArchLinux.

> without the non-free pak there is no GUI because of the font, if I'm not mistaken

Arf. Finding a FOS font seems to be the first step. Right ?

Mailaender:
As the license states you are not permitted to redistribute separately although for Linux packages you make an exemption that was too vague and insecure for me. Before risking an official deletion request, I decided to take action myself and submitted it to the http://packman.links2linux.de/ repository now where legally unclear or problematic packages are put. https://pmbs.links2linux.de/request/show/3296

PR #589, opened 2017-02-28
Fix the build for clang as it s C++11 compliant too.

devnexen:


devnexen:
I just realised that angelscript 2.31.2 version contains the fix.

PR #594, opened 2017-03-03
Identify *BSD for the startup screen

devnexen:


noway:
Nice! detectversion.py also will get this info!

yvt:
@noway Actually, that function is only used to change the caption of this button (Open Config Folder in Finder/Explorer) depending on the operating system it's running on...
screen shot 2017-03-04 at 22 27 42

`detectversion.py` is already getting an info on BSD operating systems (see [Network Protocols](https://github.com/yvt/openspades/wiki/Network-Protocols#34-version-send-c2s))

Issue #633, opened 2017-07-08
config to increase hit particles


NotAFile:
A number of users regularly complain about OpenSpades feeling less "satisfying" than AoS, eg #575. I disagree, but do think OpenSpades doesn't feel as good in some situations.

One thing I noticed while comparing AoS and OS is that hits in AoS feel a lot more substantial. I thought this might be caused by the more visible hit particles in AoS, so I changed them from 10 to a 1000 in `Client/LocalEnts.cpp:146`. While 1000 might be too much and even lower values are not "realistic", I still think it's a big improvement.

Hits can be seen from fog distance, especially at other players. Seeing the particles feels much better than a little blip on the hit indicator and you notice hits on and from your buddies a lot more.

Performance wise, this increase seems to make no difference on my intel HD graphics 4600, but if it is a performance consideration, adding an option might be good.

Issue #575, opened 2017-02-16
Feeling


Gemynal:
OpenSpades is certainly more good-looking than the Classic client (water is awesome, there is a lot of effects improving the appearance of the game) but it does not provide the same in-play feeling as Classic, which is the reason why I got back to Classic. But I am posting here because someone told me to, in fact I believe that the problem is more about the nature of the game than anything else, One reason could be that in Classic we aim with a .bmp whereas in OS we aim with a .kv6
That could also be due to the numerous animations and stuff.. So Im not hoping for a change since this is how the game work but Im just reporting a problem : Classic gameplay feels a lot better that OS gameplay, you could feel it by comparing both of them

NotAFile:
This is a very complicated issue and it is hard to describe. Every creator has a unique artistic style and it is obvious that a different creator will create a different "feeling" game.

It is very hard (especially for non-designers like us) to describe exactly what this "feeling" is, as the mood is a very subtle thing that is influenced by almost every tiny thing you see, from the color palette to the shaders and models. Game developers spend a *lot* of time changing and playing around with shaders and assets and effects to get them *just* right. My point is: making a game (or poster, or music) feel "good" is a incredibly, tremendously, unbelievably hard thing to do. It's what makes games by certain people e.g. Miyamoto so tremendously good – anybody can create a platformer, but very few will realize they need to spend months working on a swimming animation to make it feel just right, [like he did in Mario 64](http://www.gamasutra.com/view/feature/130734/game_feel_the_secret_ingredient.php?print=1). What you are requesting is hard, and in a clone game created free time with many people contributing, it is even harder.

That said, I disagree with the art direction of default OpenSpades too. I played around with the settings and now have a setup that I find very much enjoyable. I feel like the default shaders are maybe to bright and heavy – but without testing and tweaking for hours, you can't know.

Maybe you can play around with the settings and see if you find a config that feels good to you?

Gemynal:
Well I know that this is hard, but as I said Im just reporting something, not hoping for any change. I posted it because kmsi asked me to (http://aloha.pk/index.php?topic=9180.30)
Otherwise, it might be because of my bad computer too
But I disagree a bit with what you said, I got the thing about tiny details, shader, palette etc but thats not my problem, it is more about the control feeling, the moves etc. Classic feels more "direct" to me, more "tight", more "accurate"
this feeling is very hard to describe since my mother tounge is french and not english, but these words seem to be the most accurate ones to describe what I feel

[...]

PR #648, opened 2017-09-03
Add a headshot bang sound like in classic

teodor6140:
The audio file was recreated by me and doesn't sound as good as the original one
include the original in a mod pack?


teodor6140:
Related to #633 and #575

NotAFile:
Huh, I never noticed this sound existing and can't remember hearing it. Interesting.

Clogal:
Sounds good!

yvt:
Thanks for PR!

`whack.wav` is the same sound as the one when you hit someone with a spade. Neither do I remember hearing it as a headshot feedback. Even in this case, maybe it sounds a little bit blatant?

NotAFile:
So, I just checked vanilla and yes it exists, but it's very hard to hear from more than a few blocks away.

NotAFile:
Here's a video demonstrating it

https://youtu.be/OKzyoaUFYNA

Clogal:
Yes! in classic its kinda hard to hear, maybe lowing the volume a bit to match beta better

teodor6140:
Lowered the volume from 50 to 30

Sources/Client/Client_Update.cpp
```diff
@@ -936,6 +936,12 @@ namespace spades {
 					Handle<IAudioChunk> c =
 					  audioDevice->RegisterSound("Sounds/Weapons/Spade/HitPlayer.opus");
 					audioDevice->Play(c, hitPos, AudioParam());
+				} else if(type == HitTypeHead) {
+					Handle<IAudioChunk> c =
+					  audioDevice->RegisterSound("Sounds/Weapons/Impacts/whack.wav");
+					AudioParam param;
+					param.volume = 30.f;
+					audioDevice->Play(c, hitPos, param);
```
yvt: I suggest using `PlayLocal(c, param)` instead of `Play(c, hitPos, param)` since it is a sound feedback provided as a reward for a successful headshot, and is not a sound caused by others' actions. In this case, you might want to lower the volume to less than `1` since it won't be no longer attenuated by the distance.

NotAFile: I don't think playing it locally is a good idea, exactly because of the fact that it's not attenuated by distance, since it is attenuated by distance in AoS

teodor6140: I also don't remember hearing the sound when a player is getting shot not by me
I didn't use PlayLocal because as @NotAFile said volume doesn't change with distance and it sounds weird

NotAFile:
After trying this today, I must say I'm really not a fan of the way it sounds currently. It feels out of place and is way to bright.

PR #653, opened 2017-09-07
Skip dead players when spectating between spawns

NotAFile:
This makes e.g. spectating the last player alive in arena easier


yvt:
Can you add a config variable (by using `DEFINE_SPADES_SETTING`) so users can toggle this behavior?

NotAFile:
I don't think there is a good reason to spectate dead players unless you are on the spectator team, but if you require this I'll add it. Most people will not find this setting, but I guess if somebody complains you can tell them to disable it then.

yvt:
That's the point of using config variables. Most of them aren't exposed on GUI and sometimes aren't even documented, but serve very specific needs.

PR #654, opened 2017-09-07
WIP: Show Dead players in gray on scoreboard

NotAFile:
This is done for all cases, but that is probably a bad idea. While this
does not matter in public games, in small tournaments this could be an
advantage. Ideally, the server could disable giving players this
information.

Another option is to enable it only above a certain player count or only on arena mode for now.

### DIFF
--- a/Sources/Client/ScoreboardView.cpp
+++ b/Sources/Client/ScoreboardView.cpp
@@ -49,7 +49,7 @@ namespace spades {
 		static const Vector4 spectatorTextColor = {220.f / 255, 220.f / 255, 0,
 		                                           1}; // Goldish yellow
 		static const auto spectatorTeamId = 255;       // Spectators have a team id of 255
-		
+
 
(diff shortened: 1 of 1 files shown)


NotAFile:
Well, this actually isn't WIP. But You might want to request certain changes.

NotAFile:
Hm, I'm not sure you actually want to merge it without condition like it actually is right now, as it gives players more information. My suggestion would be to only implement it if spectator or more than 12 players are on the server. 6 players is the point I figure it will be hard to keep track of enemy players.

yvt:
I assumed everyone has agreed upon merging this patch at this state. Well then, yeah, this should be feature-gated depending on some conditions. I’m actually for controlling the behavior via a server-side flag passed by a newly introduced packet type.

NotAFile:
Sure, that would be good, but that's still a bit far out, I'd rather strictly feature-gate it now and then allow setting it on the server later.

PR #659, opened 2017-09-10
Adjust the gain of hit feedback sound

### LINKED ISSUE #652: Make helmet hit sound more subtle
#648 added a helmet hit sound, but I feel it's currently not subtle enough and does not fit the mood.

Before this goes into a full release, I feel that it should be made lower and perhaps a little more quiet.

Otherwise I'm really happy with the effect, it's scary to stand somewhere and have a mate next to you get dinged in the head, achieving a similar effect to #633.

yvt:
Fixes #652.

The sound gain is now constant and does not change regardless of the distance. This imitates the behavior often seen in other first-person shooters. The default gain is adjusted to the level that I feel most comfortable.

Switching from `Play` to `PlayLocal` only affects the way the gain is computed since `HeadshotFeedback.opus` is a stereo sound file and therefore not spatialized by any known audio backends.

This also introduces a config variable named `cg_hitFeedbackSoundGain` that allows users to adjust the volume of the hit feedback sound.

### DIFF
--- a/Sources/Client/Client_Update.cpp
+++ b/Sources/Client/Client_Update.cpp
@@ -51,6 +51,7 @@
 DEFINE_SPADES_SETTING(cg_ragdoll, "1");
 SPADES_SETTING(cg_blood);
 DEFINE_SPADES_SETTING(cg_ejectBrass, "1");
+DEFINE_SPADES_SETTING(cg_hitFeedbackSoundGain, "0.2");
 
 SPADES_SETTING(cg_alerts);
 SPADES_SETTING(cg_centerMessage);
@@ -962,8 +963,8 @@ namespace spades {
 					Handle<IAudioChunk> c =
 					  audioDevice->RegisterSound("Sounds/Feedback/HeadshotFeedback.opus");
 					AudioParam param;
-					param.volume = 10.f;
-					audioDevice->Play(c, hitPos, param);
+					param.volume = cg_hitFeedbackSoundGain;
+					audioDevice->PlayLocal(c, param);
 				}
 				
 				hitFeedbackIconState = 1.f;


NotAFile:
I still disagree with this, since the sound was never meant to be a hit feedback sound. It is supposed to be the sound of a bullet hitting the helmet in Voxlap, and played instead of the body hit sound, this is why it scales with distance in Voxlap.

yvt:
I don't really see the reason we have to adhere to the original client's subtle behavior so strictly. The original client and OpenSpades are different in many ways and never the same. We have a freedom of making changes as long as they don't provide an unfair advantage to the players. These differences somehow attracted many players to this game.

I merged #648 because it seemed we agreed that the introduction of the headshot feedback sound was necessary to provide a better user experience. However, strictly copying the behavior of our beloved game doesn't mean providing the *best* experience. In fact, I can't think of even one game where hit feedback sounds (not just bing but also including flesh impact sounds, but excluding explosions) decay by distance.

It's a design matter, basically. So far, I've designed this game with a goal of resembling the feel of modern games (from my view point). First, I decided to include the headshot sound because of the following reasons:

1. I felt satisfied when it was given as a reward for a successful headshot and it was [cool](http://tvtropes.org/pmwiki/pmwiki.php/Main/RuleOfCool).
2. I had been constantly annoyed when a seemingly successful headshot didn't result in an instakill because I couldn't tell if it was actually a body shot or it was rejected by the server for some reason, but now I can tell.

And then I decided to make the feedback's gain constant because of the following reasons: 

1. To follow the conventions of some other popular games.
2. The headshot sound is always overlaid with a local gun shot sound, so its volume should be kept in a certain range to maintain its effectiveness. If it's too loud they'll turn it off. If it's too subtle they don't feel rewarded.
3. Sound effects doesn't have to be an accurate simulation of the reality. In fact, gun sounds of most games are designed to be "cool", not realistic.
4. Suppose we followed 2 and and the same time adjusted the gain so it is decayed by distance. The range of gain would be so narrow that players wouldn't notice.
5. A constant gain is easy to configure because you just need one number.

On the other hand, I am still not very certain with entire the thing because:

1. I'm worried it might sound weird to the existing OpenSpades players, whether the gain constant or not. I guess new players will recognize this as a feedback and [get used to it](https://en.wikipedia.org/wiki/Suspension_of_disbelief) immediately.
2. None of games I know have a clear headshot sound like this. Some games provide a clear sound when you hit an armor, though.

I'm still open to opinions, which is why I opened this PR instead of making changes directly.

By the way, you can use `AudioParam::referenceDistance` to make sounds not get louder within that distance. It defaults to `1.0`.

evorios:
Where I might download _"HeadshotFeedback.opus"_?

PR #668, opened 2017-10-14
On OpenBSD SDL_GetBasePath is not really implemented

devnexen:
due to lack of kernel support of exe self-retrieval.


NotAFile:
Does this apply to all *BSDs?

devnexen:
Nope, FreeBSD works just fine.

PR #683, opened 2017-12-10
Do no overwrite the User Resource Directory on every snap launch

### LINKED ISSUE #667: snap does not remember config
So i installed openspades with
`sudo snap install openspades`

and it works fine. Except one thing:
Whenever I re-open the game, the settings reset. E.g: the settings at the start menu screen, it always asks for a username

Ubuntu 17.04

feikname:
Hopefully fixes #667.

### DIFF
--- a/snap/scripts/launch_openspades
+++ b/snap/scripts/launch_openspades
@@ -10,15 +10,6 @@ case "$SNAP_ARCH" in
 	;;
 esac
 
-if [ ! -f $SNAP_USER_DATA/.local/share/openspades/Resources ];
-then
-  mkdir -p $SNAP_USER_DATA/.local/share/openspades/Resources
-	cp $SNAP/share/games/openspades/Resources/* $SNAP_USER_DATA/.local/share/openspades/Resources
-	echo cl_checkForUpdates: 1 > $SNAP_USER_DATA/.local/share/openspades/Resources/SPConfig.cfg
-	echo s_eax: 1 >> $SNAP_USER_DATA/.local/share/openspades/Resources/SPConfig.cfg
-fi
-
-
 export LD_LIBRARY_PATH=$SNAP/usr/lib/$ARCH:$LD_LIBRARY_PATH
 
 # XKB config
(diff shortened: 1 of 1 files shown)


snap/scripts/launch_openspades
```diff
@@ -10,15 +10,6 @@ case "$SNAP_ARCH" in
 	;;
 esac
 
-if [ ! -f $SNAP_USER_DATA/.local/share/openspades/Resources ];
```
yvt: Is removing this entire `if`-`fi` block necessary? Replacing `-f` with `-d` should at least fulfill the goal, I guess.

feikname: You're right, replacing `-f` with -`d` should be enough fix this. But I believe this whole block is not necessary so the commit removes it altogether.

What the current script does:
  1. Check for the existence of the User Resource Directory, and if it doesn't: (this step is broken due to wrong `-f` flag and is ran everytime OpenSpades launchs)
  2. Copies the Global Resource Directory to the User Directory, which seems useless to me. Maybe it's good to simplify modding?
  3. Creates an initial config file (that I'm unsure if gets overwritten on OpenSpades on first launch or not, I could test this on an normal OpenSpades build), which:
  4. Enables OpenSpades to automatically search for updates instead of letting this be an one time opt-in.
 5. Enables OpenAL by default (assuming 1 means enable in the config option). I think this is a mistake by the author, because his intention was clearly to have it disabled by default because it crashes OS on 32-bit linux apparently.

Regarding point 2, I think that's totally useless.\
Regarding point 3, I'm almost it does get loaded by OpenSpades on first launch\
Regarding point 4, the author said something about it being broken, we should probably take a look at that. I still think this line should be removed.\
Regarding point 5, I think we should let this option be chosen by OpenSpades and just put a warning in the README and wiki about 32-bit snaps enabling OpenAL.

To be honest, I couldn't test the changes myself due to some weird dynamic library issues when trying to run the developer  snapcraft I built.

To quote the author: (https://popey.com/blog/posts/openspades-snap-pew-pew.html)
> The only slightly kludgy thing I did was copy the Resources directory into a writable directory and modify the default configuration for first-launch. OpenSpades has an 'update check' feature which looks for game updates online, but it seems to me this doesn't work on Linux - probably due to there being no binary builds currently available. So rather than the user getting prompted to enable this feature, I enabled it by default, so the user just gets a notification that they're on the latest version, rather than being asked to enable a feature that won't work anyway. With snaps the user will get an update as soon as it's pushed to the store, so no need for the in-app update mechanism I think.
>
> I also had to disable EAX (positional audio) because it crashes OpenAL on the i386 builds. Players on AMD64 systems can of course just re-enable this, as it's only the default I set, not permanently off. I might modify this default so it's only disabled on i386 builds in future.

@popey what do you think?

feikname:
I will close this in favor of a simpler pull request for only solving #667

PR #692, opened 2017-12-27
Those methods can t be purely virtual as the class itself call them.

devnexen:


theunknownxy:
It's correct that the methods are purely virtual. The concrete implementations are in subclasses (in this case `StandardUpdateFeed` contains an implementation) and it doesn't make sense to instantiate a `HttpUpdateFeed` directly. For this reason it makes sense that they are *purely* virtual, causing the class to be abstract.
(If you don't know why this whole thing works, you might want to read a bit about object-orientation.)

devnexen:
In fact clang consider the calls from Run() being virtual method calls ... Could be compiler bug, but it led to a segfault when quitting in my case.

theunknownxy:
Apparently calling a pure virtual function from a destructor is [undefined](http://en.cppreference.com/w/cpp/language/abstract_class) behaviour. It could be a race condition of the thread calling the method and the join in the destructor. I guess your fix relies on the mechanism how the vtables are implemented in your compiler. 
Oh... the joys of undefined behaviour!

devnexen:
@yvt any opinion ? :-)

devnexen:
I ve tested those :
- OpenBSD/GCC no issues.
- FreeBSD/clang no issues.
- OpenBSD/clang the issue.

It seems more to me specific compiler issue.

PR #694, opened 2017-12-29
Add color coding to ping

NotAFile:
Currently, the following steps are used:
 * <90: green
 * <120: yellow
 * <200: orange
 * >200: red

It currently looks waaay to colorful. I think a colored bar or something might
look best, but for now this works and lets you quickly see how good you can
expect the connection to be


yvt:
LGTM

PR #707, opened 2018-01-05
Add Counter for world updates per Second

NotAFile:
This work was just done for personal testing and is probably not up to the
required code standards. Nonetheless, I have decided to send a PR as it is very
useful information and a good starting point.

### DIFF
--- a/Sources/Client/Client.h
+++ b/Sources/Client/Client.h
@@ -99,6 +99,7 @@ namespace spades {
 			};
 
 			FPSCounter fpsCounter;
+			//FPSCounter upsCounter;
 
 			std::unique_ptr<NetClient> net;
 			std::string playerName;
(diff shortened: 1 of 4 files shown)


Sources/Client/Client.h
```diff
@@ -265,16 +266,16 @@ namespace spades {
...
+			 *
+			 * Throws an exception if the current camera mode does not have a player in concern.
 			 */
 			int GetCameraTargetPlayerId();
 
 			/**
 			 * Retrieves the target player of the current camera mode (as returned by
 			 * `GetCameraMode`).
-		     *
-		     * Throws an exception if the current camera mode does not have a player in concern.
```
NotAFile: These used to be spaces so I changed them

Sources/Client/Client.h
```diff
@@ -99,6 +99,7 @@ namespace spades {
 			};
 
 			FPSCounter fpsCounter;
+			//FPSCounter upsCounter;
```
yvt: This line is obviously redundant.

Sources/Client/Client.h
```diff
@@ -395,6 +396,7 @@ namespace spades {
 			virtual ~Client();
 
 		public:
+			FPSCounter upsCounter;
```
yvt: Don’t add it as a public member, please. I suggest to move it to `NetClient` and then expose the ups value via a public method.

NotAFile: Yes, that was the original plan. But then I got an error about it being incomplete (NetClient and Client would have a cyclic dependency?) and gave up because I basically only quickly wanted this information once.

I'd probably have to brush up on my C++ to figure out how to deal with that.

NotAFile: Would adding a MarkWorldUpdate() be better? That's the other idea I had to bring this up to OpenSpades code quality.

yvt: Yep, that’d be better.

PR #766, opened 2018-09-01
gunplay should match voxlap now

BR-:
recoil, spread, and damage values are identical
recoil algorithm should match as well

notably, the shotgun is SIGNIFICANTLY weaker after this commit - on average it will do half as much damage as before

This is correct to the best of my knowledge, but it would be nice if people could doublecheck. I've been playing on it for a while and it feels right.
I haven't played 0.76 very much, I don't know how it should feel, but I'm pretty sure this will match it as well.

### DIFF
--- a/Sources/Client/Player.cpp
+++ b/Sources/Client/Player.cpp
@@ -556,9 +556,8 @@ namespace spades {
 			float spread = weapon->GetSpread();
 			GameMap *map = world->GetMap();
 
-			if (weapInput.secondary) {
-				// vanilla behavior (confirmed by measurement)
-				spread *= 0.5f;
+			if (!weapInput.secondary) {
+				spread *= 2;
 			}
 
 			// pyspades takes destroying more than one block as a
--- a/Sources/Client/Player.cpp
+++ b/Sources/Client/Player.cpp
@@ -762,18 +762,37 @@ namespace spades {
 			Vector3 rec = weapon->GetRecoil();
 			float upLimit = Vector3::Dot(GetFront2D(), o);
 			upLimit -= 0.03f; // ???
-			o += GetUp() * std::min(rec.y, std::max(0.f, upLimit)) *
-				(input.crouch ? 0.5f : 1.0f);
-			// vanilla's horizontial recoil seems to driven by a triangular wave generator.
-			// the period was measured with SMG
-			float triWave = world->GetTime() * 0.9788f;
-			triWave -= std::floor(triWave);
-			if (triWave < 0.5f) {
-				triWave = triWave * 4.0f - 1.0f;
+
+			// vanilla's horizontial recoil is driven by a triangular wave generator.
+			int time = (int) (world->GetTime() * 1000);
+			float triWave;
+			if (time % 1024 < 512) {
+				triWave = (time % 512) - 255.5;
 			} else {
-				triWave = 3.0f - triWave * 4.0f;
+				triWave = 255.5 - (time % 512);
 			}
-			o += GetRight() * rec.x * triWave * (input.crouch ? 0.5f : 1.0f);
+
+			float horzModifier = 1;
+			float vertModifier = 1;
+
+			if ((input.moveLeft || input.moveRight || input.moveForward || input.moveBackward) && !weapInput.secondary) {
+				horzModifier *= 2;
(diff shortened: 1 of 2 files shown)


feikname:
I added the `needs-confirmation` tag because I believe at least one person (preferably three) should double check this before the merge.

10se1ucgo:
Seems to match up with what I and mat's client have, except the spread looks about twice as large as it should be. But maybe that's just my misunderstanding of how OS calculates spread

NotAFile:
from memory, last time we concluded that OS calculates spread and knockback differently

NotAFile:
I also think it's worth deliberately not nerfing the shotgun, because the voxlap shotgun is just too useless.

BR-:
The reason I didn't send this earlier was that I couldn't decide what to do with the shotgun. Some people have complained that OS makes shotguns too strong in arena - not sure how much I believe it. I think 
OS shotgun is usable even on larger maps, but at least to me the variance is too high to be strong. Some hit test debug images from OS shotgun, if you haven't used it much, to get a feel for it: https://imgur.com/a/JkkKPux

OpenSpades and ace/classic raw spread values should be off by 2x - OS does "0.5x if scoped" and the others do "2x if unscoped":
https://github.com/10se1ucgo/ace/blob/61ea8cc6c9a72579a7c2e7ee8e0c52bf1571344a/src/weapon.cpp#L417
https://github.com/yvt/openspades/blob/master/Sources/Client/Player.cpp#L559
I'll change the pullreq to match.

The classic horizontal wave is kind of dumb for non-SMG recoil. The period is 1.024s and the rifle shoots twice in 1s, meaning if you're shooting as fast as possible, every other shot will have almost the same recoil (and the others will be almost the same but mirrored). The shotgun shoots once per 1s, so it's even more of a problem. This pullreq should perfectly match classic, but we should think about if they should just be normal RNG. (this maybe should be a separate issue)

The horizontal wave period change also changed the multiplier for horizontal recoil values by 255.5x. The new 0.75 values, after scaling to match the old ones:
Rifle, 0.025 before this pullreq, 0.02555 after.
SMG, 0.01 before, 0.012775 after.
Shotgun, 0.05 before, 0.0511 after.

Notably, the existing 0.76 values match the classic client exactly - I assume greaser looked at it when adding 0.76 compatibility. Unfortunately, he didn't look at how the calculations worked, so they ended up with 255.5x less horizontal recoil and 2x less spread than they should have had. This pullreq is a big nerf to 0.76 OS.

NotAFile:
can we work on merging this after re-buffing the shotgun?

Sonicscream:
@yvt take a look at this?
Currently the SMG on 0.76 OpenSpades has absolutely no horizontal recoil. Without this patch, it makes it unfair to other clients running on 0.76.

Clogal:
I don't think shotgun needs a buff, at least in voxlap if you master it, it could go pretty decent, but now it takes zero skill

iamgreaser:
As the original author of the PR which added 0.76 support in the first place, I can confirm that the current OpenSpades X recoil values for 0.76 are completely wrong when fed through the current OpenSpades recoil logic, but are completely right when fed through the vanilla client's logic. This was made apparent when we attempted to test 0.76 support in piqueserver and people noticed a complete lack of X recoil on the SMG.

I am not in a particularly good position to prove that the implementation is perfect, but it's definitely a huge improvement over what we've got right now so that's a :+1: from me.

The only concern I have with this patch is that, while the SMG is using the correct delay value as per the vanilla client, I'm not sure if the rest of the engine is set up to emulate the slightly wonky timing in that department... assuming the timing is still wonky in 0.75.

Still, this is a huge improvement in the 0.76 department, and without this PR, we can't really run 0.76 servers without banning OpenSpades clients from them, and that would suck because OpenSpades is quite possibly the most popular client and also a key part in actually getting this game thriving again.

----

And as for that comment about not nerfing the shotgun, I do admit I've been enjoying what the shotgun's at right now and I would be in favour of having that as a base for cleaning up the weapon balance via a protocol extension and then eventually a new protocol version once people are happy with the results. But for now, I'm going to have to enjoy the pain of the 0.75 shotgun being even more of a 0.75ly meme.

yvt:
Sorry for the delay. I'll merge it soon!

PR #770, opened 2018-09-02
modern gcc and clang build fix

devnexen:
putting constraints on g rather than r as well.


NotAFile:
Perhaps the angelscript changes should be upstreamed instead?

theunknownxy:
They are fixed upstream:
 * https://sourceforge.net/p/angelscript/code/2410/
 * https://sourceforge.net/p/angelscript/code/2335/

devnexen:
Yes basically it is a bit of backporting here, the change from r to g constraint is mainly to cope with modern sanitizers rather than anything else.

PR #774, opened 2018-09-17
Improve/fix random number generation

yvt:
The ~~global~~ RNG states are now protected from simultaneous accesses.

~~Performance sensitive code requiring generation of a large number of random numbers can utilize `LocalRNG`. This is not thread-safe but provides a much greater throughput. `LocalRNG` is seeded by the global RNG only at construction time.~~

In summary, every random number generation is done via the following method ~~either of the following methods~~:

- The `SampleRandom*` global function family providing an access to a thread-local random number generator ~~a global random number generator instance shared by all threads~~.
- ~~`LocalRNG`, which is a high-throughput, low-quality random number generator, initialized using the global RNG.~~

## Tasks

- [x] Check if it builds with Clang
- [x] Check if it builds with MSVC 2017
- [x] Check if it builds with GCC 4.9

### DIFF
--- a/Sources/Core/Math.h
+++ b/Sources/Core/Math.h
@@ -53,7 +53,7 @@ namespace spades {
 		auto begin = std::begin(container);
 		auto end = std::end(container);
 		auto numElements = std::distance(begin, end);
-		begin += SampleRandomInt<typeof(numElements)>(0, numElements - 1);
+		begin += SampleRandomInt<decltype(numElements)>(0, numElements - 1);
 		return *begin;
 	}
 
--- a/Sources/Core/Math.h
+++ b/Sources/Core/Math.h
@@ -42,6 +42,9 @@ namespace spades {
 	/**
 	 * Generates an integer in a specified inclusive range.
 	 * This function is thread-safe.
+	 *
+	 * `T` must be one of `int`, `unsigned int`, `long`, `unsigned long`,
+	 * `short`, `unsigned short`, `long long`, and `unsigned long long`.
 	 */
 	template <class T = int> T SampleRandomInt(T a, T b);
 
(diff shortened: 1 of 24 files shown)


Sources/Core/Math.h
```diff
@@ -27,12 +27,92 @@
...
+
+	/** Generates a random `bool`. This function is thread-safe. */
+	inline bool SampleRandomBool() { return globalThreadSafeRNG() & 0x1; }
+
+	/** Get a mutable reference to a random element from a container. */
+	template <class T> inline typename T::reference SampleRandomElement(T &container) {
+		auto begin = std::begin(container);
+		auto end = std::end(container);
+		auto numElements = std::distance(begin, end);
+		begin += SampleRandomInt<typeof(numElements)>(0, numElements - 1);
```
theunknownxy: `typeof(...)` seems to be some GCC extension and I'm not sure whether this will compile on MSVC. Fortunately there is `decltype(...)` since C++11.

Sources/Core/Math.h
```diff
@@ -27,12 +27,92 @@
...
+		ThreadSafeRNG();
+
+		result_type operator()();
+
+		static constexpr result_type min() { return inner_type::min(); }
+		static constexpr result_type max() { return inner_type::max(); }
+
+	private:
+		inner_type inner;
+		std::mutex mutex;
```
theunknownxy: Just an idea: what about using an thread-local lazily constructed rng? (Inspired by rust's [rand](https://docs.rs/rand/0.5.5/rand/fn.thread_rng.html) crate).

Sources/Core/Math.h
yvt: Ah, that somehow slipped my mind probably because I haven’t worked with C++ for so long time. Thanks for pointing that out.

Sources/Core/Math.h
yvt: That’s definitely a better idea. (Actually I thought of that at some point but I was about to sleep so I kept going on)

Issue #27, opened 2013-08-29
Jumping and reloading while sprinting.


AEM7:
You can jump and reload while sprinting when this not possible for regular users.

Issue #26, opened 2013-08-29
Blocks are supposed to break in 2 hits, not 3.


AEM7:
Blocks are supposed to break in 2 hits, not 3.

Issue #25, opened 2013-08-29
cannot conntect/get kicked


Scipio-Aloha:
When i try to play on aloha or anyother compatible servers i can play then i get a disconnection error

Thu Aug 29 06:41:45 2013 [DirectoryFileSystem.cpp:28] Directory File System Initialized: F:\Documents and Settings\Games\Desktop\OpenSpades-0.0.7-Windows\Resources (Read-only)
Thu Aug 29 06:41:45 2013 [DirectoryFileSystem.cpp:28] Directory File System Initialized: F:\Documents and Settings\Games\Application Data\OpenSpades\Resources (Read/Write)
Thu Aug 29 06:41:45 2013 [Main.cpp:97] Log Started.
Thu Aug 29 06:41:45 2013 [Main.cpp:127] Pak Registered: Base.pak

Thu Aug 29 06:41:45 2013 [Main.cpp:127] Pak Registered: Models.pak

Thu Aug 29 06:41:45 2013 [Main.cpp:127] Pak Registered: Sounds.pak

Thu Aug 29 06:41:45 2013 [MainWindowHandler.cpp:91] Loading Preferences to MainWindow
Thu Aug 29 06:41:45 2013 [MainWindowHandler.cpp:108] Video Mode Found: 1920x1080
Thu Aug 29 06:41:45 2013 [MainWindowHandler.cpp:108] Video Mode Found: 1680x1050
Thu Aug 29 06:41:45 2013 [MainWindowHandler.cpp:108] Video Mode Found: 1600x900
Thu Aug 29 06:41:45 2013 [MainWindowHandler.cpp:108] Video Mode Found: 1440x900
Thu Aug 29 06:41:45 2013 [MainWindowHandler.cpp:108] Video Mode Found: 1280x1024
Thu Aug 29 06:41:45 2013 [MainWindowHandler.cpp:108] Video Mode Found: 1280x960
Thu Aug 29 06:41:45 2013 [MainWindowHandler.cpp:108] Video Mode Found: 1280x800
Thu Aug 29 06:41:45 2013 [MainWindowHandler.cpp:108] Video Mode Found: 1024x768
Thu Aug 29 06:41:45 2013 [MainWindowHandler.cpp:108] Video Mode Found: 800x600
Thu Aug 29 06:41:45 2013 [Main.cpp:140] Entering FLTK main loop
Thu Aug 29 06:41:48 2013 [SDLGLDevice.cpp:99] SDLGLDevice created
[...]

Issue #24, opened 2013-08-29
0.76 support needs a drop-down menu


iamgreaser:
0.76 is supported but it's tucked away in the Advanced Settings section (cg_protocolVersion). Perhaps a Version Selection dropdown would be good?

feikname:
Currently, in 0.012b and the latest build of 0.1.0, there is a toggle for that, it works similiar to the dropdown suggestion. You can select for 0.75 only, 0.76 only, or both.

PR #775, opened 2018-10-06
surface lock issue with ref counter.

devnexen:
both at object destruction leads to use after free so
we store the information at creation time instead.

### DIFF
--- a/Sources/Gui/SDLRunner.cpp
+++ b/Sources/Gui/SDLRunner.cpp
@@ -314,6 +314,7 @@ namespace spades {
 		class SDLSWPort : public draw::SWPort {
 			SDL_Window *wnd;
 			SDL_Surface *surface;
+			bool locked;
 			bool adjusted;
 			int actualW, actualH;
 
@@ -330,7 +331,7 @@ namespace spades {
 
 		protected:
 			virtual ~SDLSWPort() {
-				if (surface && SDL_MUSTLOCK(surface)) {
+				if (surface && locked) {
 					SDL_UnlockSurface(surface);
 				}
 			}
@@ -339,7 +340,7 @@ namespace spades {
 			SDLSWPort(SDL_Window *wnd) : wnd(wnd), surface(nullptr) {
 				surface = SDL_GetWindowSurface(wnd);
 				// FIXME: check pixel format
-				if (SDL_MUSTLOCK(surface)) {
+				if ((locked = SDL_MUSTLOCK(surface))) {
 					SDL_LockSurface(surface);
 				}
 				actualW = surface->w & ~7;


yvt:
Would you care to elaborate on the issue? I assumed that `SDL_MUSTLOCK` is a pure function (i.e., the output is constant for a specific input), so there doesn’t seem to be a point in remembering the value of `SDL_MUSTLOCK(surface)`.

devnexen:
You can check with a sanitiser, at exit the game crashes because of the reference counting.

yvt:
No issues detected by AddressSanitizer on my machine.

devnexen:
Did you use gcc's or llvm's one ? used clang 7

yvt:
I tried clang's one via Xcode 9.4's build options. (Apple uses a different versioning scheme on their custom LLVM toolsets, so their version numbers don't directly translate to upstream)

```
clang --version
Apple LLVM version 9.1.0 (clang-902.0.39.2)
Target: x86_64-apple-darwin17.7.0
```

devnexen:
Yes clang from xcode is behind the current last release, maybe clang/llvm form homebrew has different outcome, I tried under Linux.

yvt:
Tried with Clang 6.0.1 (on x86_64 Linux) and got nothing useful. Maybe can you just post the output of sanitizer?

devnexen:
screenshot_2018-10-09_07-03-58

theunknownxy:
I can't reproduce it neither with GCC 8.2.0/Clang 7.0.0's AddressSanitizer nor with Valgrind on a standard debug build (All three on Linux).
But something seems to be off in the sanitizer output you posted: The code accessed memory which was freed by something unrelated to SDLSWPort. Either it's broken in a really strange way or the sanitizer hasn't resolved the symbols correctly. But the output is cut off at the end, so it's hard to say which one is the case.

But in the case that the sanitizer did resolve the symbols correctly, I have a theory about what's happening: According to the documentation [SDL_GetWindowSurface](https://wiki.libsdl.org/SDL_GetWindowSurface) returns a surface which is invalidated after resizing and you need to call SDL_GetWindowSurface again. I haven't seen anything which calls this function except the constructor of SDLSWPort. (Yeah the window is not resizable... except some window manager might ignore that hint and do it anyway). When the initial surface is freed, the region might be reused by something else (like c2t). This writes some data into the region and some time later frees it again. But now c2t is marked as the previous owner of the memory region and shows up in the backtrace. Finally when the SDL_MUSTLOCK check is executed, it accesses unallocated memory, which in turn *might* trigger an segmentation fault or an sanitizer warning.
The reason why your change would "fix" the problem in this case, is that the surface probably won't need locking and thus nothing which needs the intiial SDL_Surface structure is executed anymore.

devnexen:
I will post later then lower end of the trace if interested then.

yvt:
I think I figured out the problem. Here's the explanation:

The following part of the backtrace indicates that a `FTFont` was destroyed during a gameplay.

> ...
> ```
>     #10 0xa064c3 in spades::ngclient::FTFont::~FTFont() (/home/dcarlier/Contribs/openspades/Build/bin/openspades+0xa064c3)
>     #11 0xa065e8 in spades::ngclient::FTFont::~FTFont() (/home/dcarlier/Contribs/openspades/Build/bin/openspades+0xa065e8)
>     #12 0xcc82f8 in spades::RefCountedObject::Release() (/home/dcarlier/Contribs/openspades/Build/bin/openspades+0xcc82f8)
>     #13 0x1490b81 in asCScriptEngine::CallObjectMethod(void*, asSSystemFunctionInterface*, asCScriptFunction*) const (/home/dcarlier/Contribs/openspades/Build/bin/openspades+0x1490b81)
>     #14 0x14905d9 in asCScriptEngine::CallObjectMethod(void*, int) const (/home/dcarlier/Contribs/openspades/Build/bin/openspades+0x14905d9)
>     #15 0x14e97cd in asCScriptObject::ReleaseAllHandles(asIScriptEngine*) (/home/dcarlier/Contribs/openspades/Build/bin/openspades+0x14e97cd)
>     #16 0x1466601 in asCScriptEngine::CallObjectMethod(void*, void*, asSSystemFunctionInterface*, asCScriptFunction*) const (/home/dcarlier/Contribs/openspades/Build/bin/openspades+0x1466601)
>     #17 0x149355c in asCScriptEngine::CallObjectMethod(void*, void*, int) const (/home/dcarlier/Contribs/openspades/Build/bin/openspades+0x149355c)
>     #18 0x143bc3c in asCGarbageCollector::IdentifyGarbageWithCyclicRefs() (/home/dcarlier/Contribs/openspades/Build/bin/openspades+0x143bc3c)
>     #19 0x143e964 in asCGarbageCollector::GarbageCollect(unsigned int, unsigned int) (/home/dcarlier/Contribs/openspades/Build/bin/openspades+0x143e964)
>     #20 0x1493740 in asCScriptEngine::GarbageCollect(unsigned int, unsigned int) (/home/dcarlier/Contribs/openspades/Build/bin/openspades+0x1493740)
>     #21 0x13fb02a in asCContext::Execute() (/home/dcarlier/Contribs/openspades/Build/bin/openspades+0x13fb02a)
>     #22 0x13d2f69 in spades::ScriptContextUtils::ExecuteChecked() (/home/dcarlier/Contribs/openspades/Build/bin/openspades+0x13d2f69)
>     #23 0x13d2b2f in spades::ScriptContextHandle::ExecuteChecked() (/home/dcarlier/Contribs/openspades/Build/bin/openspades+0x13d2b2f)
>     #24 0x12341ed in spades::client::ScriptIWeaponSkin::SetReadyState(float) (/home/dcarlier/Contribs/openspades/Build/bin/openspades+0x12341ed)
>     #25 0x8e90dd in spades::client::ClientPlayer::SetSkinParameterForTool(spades::client::Player::ToolType, asIScriptObject*) (/home/dcarlier/Contribs/openspades/Build/bin/openspades+0x8e90dd)
>     #26 0x8f491b in spades::client::ClientPlayer::AddToSceneThirdPersonView() (/home/dcarlier/Contribs/openspades/Build/bin/openspades+0x8f491b)
>     #27 0x8fa46b in spades::client::ClientPlayer::AddToScene() (/home/dcarlier/Contribs/openspades/Build/bin/openspades+0x8fa46b)
> ```
> ...

Each "screen" in the game (e.g., the startup window and the main window) is created with a `FontManager`, which provides references to `FTFont`s created for the particular `IRenderer` the screen is associated with. Once created, `FontManager` and its `FTFont`s are not supposed to be destroyed until the screen is closed, so I set a breakpoint on `FTFont::~FTFont()` to figure out why that happened during a gameplay.

Here's what I've got after keeping the client open for a while:

screen shot 2018-10-11 at 23 29 39 copy

This shows that the `FTFont` in question was in fact owned by the startup window, not the main window. AngelScript implements a tracing GC, meaning it releases objects only when it decides to, and as a result the `FTFont` didn't die with its owner. `FTFont` maintains a reference to `IRenderer`, so `SWRenderer` associated with the startup window was kept alive even after the startup window, along with the `SDL_Surface` obtained by `SDL_GetWindowSurface`, were destroyed.

And then GC kicked in — all objects used by the startup window were finally released, during which the process `~SDLSWPort()` was called but at this point the `surface` was already gone because the startup window had been closed a long time ago. The result is unpredictable — apparently it didn't do any apparent harm in most cases, but the determining factors are manifold. None of us could reproduce the crash, but it might have caused some of unexplained seemingly-unrelated random crash cases that have been stacked up for years on our bug tracker.

In conclusion, it was definitely a use-after-free of `surface`, but the real solution is to make sure the contents of `surface` is not accessed after the `SDL_Window` it was created from (or more precisely, after any actions that potentially invalidate `SDL_Surface` returned by `SDL_GetWindowSurface`). For example, `SDLSWPort` could be "invalidated" by nullifying `surface` when the associated `SDL_Window` is destroyed. 

As an alternative option, we could run a full GC before destroying a `SDL_Window`, which would work because OpenSpades explicitly disables global variables in AngelScript. However, that would be considered a bad practice since it creates an implicit dependency on a particular execution sequence of the code.

devnexen:
nullyfing surface was my first approach even though I did not know the scripting language was the real root cause. But that made the situation slighty better but still crashed time to time.

PR #793, opened 2018-12-20
Fix CMake build for macOS

Ericson2314:
My distro, NixOS, does darwin builds too.


yvt:
@Ericson2314 I pushed a few adjustments to make it compile and get it fully working on macOS. Can you check that I haven't broken anything?

Ericson2314:
I can take another look. For some context, i was using Nixpkgs (https://github.com/NixOS/nixpkgs/blob/master/pkgs/games/openspades/default.nix) which, as a distro-like thing, will build dependencies in a more structured format (and also in disjoint prefixes), so my use-case is a closer to the Linux build than somebody building things by hand ad-hoc on macOS.

cmake/FindSDL2.cmake
```diff
@@ -70,19 +70,22 @@
...
+  ~/Library/Frameworks/SDL2.framework/Headers
+  /Library/Frameworks/SDL2.framework/Headers
   /usr/local/include/SDL2
   /usr/include/SDL2
   /sw # Fink
-  /opt/local # DarwinPorts
+  /opt/local/include/SDL2 # DarwinPorts
   /opt/csw # Blastwave
   /opt
+  NO_DEFAULT_PATH
```
Ericson2314: This breaks Nixpkgs/Nixos, which installs everything to separate prefixes and then adds `-isystem` and `-L` to supply them. `pkg-config` would also work, but this file doesn't use it.

Issue #674, opened 2017-11-28
OpenGL-capable graphics accelerator is unavaiable linux snap in Manjaro


rurigk:
Logs

2017/11/28 02:16:06 [Main.cpp:324] Package: OpenSpades 0.1.1
2017/11/28 02:16:06 [DirectoryFileSystem.cpp:46] Directory File System Initialized: ./Resources (Read-only)
2017/11/28 02:16:06 [DirectoryFileSystem.cpp:46] Directory File System Initialized: /share/games/openspades/Resources (Read-only)
2017/11/28 02:16:06 [Main.cpp:400] XDG_DATA_HOME is /home/rurigk/snap/openspades/10/usr/share
2017/11/28 02:16:06 [DirectoryFileSystem.cpp:46] Directory File System Initialized: /home/rurigk/snap/openspades/10/.local/share/openspades/Resources (Read/Write)
2017/11/28 02:16:06 [Main.cpp:454] Log Started.
2017/11/28 02:16:06 [FltkPreferenceImporter.cpp:82] Checking for legacy preference file: /home/rurigk/snap/openspades/10/.fltk/yvt.jp/OpenSpades.prefs
2017/11/28 02:16:06 [FltkPreferenceImporter.cpp:86] Legacy preference file wasn't found.
2017/11/28 02:16:06 [Settings.cpp:65] Loading preferences from SPConfig.cfg
2017/11/28 02:16:06 [Settings.cpp:69] SPConfig.cfg found.
2017/11/28 02:16:06 [Main.cpp:463] ---- CPU Information ----
2017/11/28 02:16:06 [Main.cpp:464] Vendor ID: GenuntelineI
2017/11/28 02:16:06 [Main.cpp:465] Brand ID:         Intel(R) Celeron(R) CPU 1000M @ 1.80GHz
2017/11/28 02:16:06 [Main.cpp:466] Supports MMX: YES
2017/11/28 02:16:06 [Main.cpp:467] Supports SSE: YES
2017/11/28 02:16:06 [Main.cpp:468] Supports SSE2: YES
2017/11/28 02:16:06 [Main.cpp:469] Supports SSE3: YES
2017/11/28 02:16:06 [Main.cpp:470] Supports SSSE3: YES
2017/11/28 02:16:06 [Main.cpp:471] Supports FMA: NO
2017/11/28 02:16:06 [Main.cpp:472] Supports AVX: NO
2017/11/28 02:16:06 [Main.cpp:473] Supports AVX2: NO
[...]

PR #806, opened 2019-01-22
Update snapcraft

popey:
This pull request updates the yaml used to generate a snap package. I tend to run this in a 16.04 Ubuntu VM, and it will generate a build based on the last stable release tag if that's newer than the one currently in the beta channel in the snap store. If the one in the snap store beta channel is already the latest tagged release then it will build whatever git master contains. This makes it easy to build a stable release of openspades. Once merged, someone from the openspades project (@yvt ?) can go to build.snapcraft.io and hook this repo up. It should build on each commit and land a release in the store. You can then use the store dashboard to release those. if you did that now, it'll build 0.1.3.

Happy to answer any questions or assist if I can :)


feikname:
Unrelated to this PR, but do you have any ideas about what might be the issue with #674?

tigercoding56:
unrelated but could you make the snap build on arm64 too ?

Issue #788, opened 2018-11-22
Release game on Flatpak/Flathub


MyNameIsJulian:
On Linux the game is currently easily downloadable using Snap, but there are a couple of issues with that. The main issue is that Snap creates a loop device for every package you install with it, so imagine installing each package on it's own USB drive and then going to disk manager, it's a mess. Second is that Snap is centralized with no option for remote repositories, which a lot of people don't really like.

Flatpak does what Snap does and does it better. A lot of popular distros like Linux Mint and Manjaro (both more popular than Ubuntu under desktop users) come with Flatpak by default and not Snap. Please release this game as a Flatpak(on Flathub or a self-hosted repo), that would be really great.

kdeleteme:
I did some [initial work](https://github.com/kdeleteme/openspades/tree/master/flatpak) on flatpak packaging. While compiling, it recognizes the dependencies but I have trouble getting it to actually compile. Would love to have help.

@yvt Would a pull request be welcome if we get this to work?

yvt:
@kdeleteme Yes, much appreciated!

theunknownxy:
I also wrote an basic builder script a while ago: [theunknownxy/openspades-flatpak](https://github.com/theunknownxy/openspades-flatpak). @kdeleteme feel free to use it as an source of inspiration.

kdeleteme:
@theunknownxy Thanks! I checked it out found the script to be really helpful.

@yvt Here's the log during compilation (I'm using clang++):
```bash
-- The C compiler identification is Clang 6.0.1
-- The CXX compiler identification is Clang 6.0.1
-- Check for working C compiler: /usr/bin/clang
-- Check for working C compiler: /usr/bin/clang -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /usr/bin/clang++
-- Check for working CXX compiler: /usr/bin/clang++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
CMake Deprecation Warning at CMakeLists.txt:16 (cmake_policy):
  The OLD behavior for policy CMP0054 will be removed from a future version
  of CMake.

  The cmake-policies(7) manual explains that the OLD behaviors of all
  policies are deprecated and that a policy should be set to OLD only under
[...]

Issue #686, opened 2017-12-15
XDG_DATA_HOME is not being taken into consideration to decide the User Resource Directory location


feikname:
Instead, `$HOME/.local/share` is **ALWAYS** used. which is the fallback directory for when XDG_DATA_HOME is empty or unset, as defined by the latest [XDG Base Directory Specification](https://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html).

Relevant output:
```
feik@name:~$ XDG_DATA_HOME=/home/feik openspades
2017/12/15 05:43:17 [Main.cpp:324] Package: OpenSpades 0.1.1
2017/12/15 05:43:17 [DirectoryFileSystem.cpp:46] Directory File System Initialized: ./Resources (Read-only)
2017/12/15 05:43:17 [DirectoryFileSystem.cpp:46] Directory File System Initialized: /usr/local/share/games/openspades/Resources (Read-only)
2017/12/15 05:43:17 [Main.cpp:400] XDG_DATA_HOME is /home/feik
2017/12/15 05:43:17 [DirectoryFileSystem.cpp:46] Directory File System Initialized: /home/feik/.local/share/openspades/Resources (Read/Write)
```

PR #811, opened 2019-02-12
Add flatpak build

kdeleteme:
#788 

I wrote a manifest to build openspades with flatpak. This may help enable distribution through [flathub](https://flathub.org) if desired.


flatpak/jp.yvt.OpenSpades.appdata.xml
```diff
@@ -0,0 +1,66 @@
...
+    first-person shooter created by Ben Aksoy, featuring fully destructible
+    terrain and plenty of game modes (including the well-known Capture the
+    Flag) created by the community.</p>
+    <ul>
+      <li>Can connect to a vanilla/pyspades/pysnip server</li>
+      <li>Uses OpenGL/AL for better experience</li>
+      <li>Open source, and cross platform</li>
+    </ul>
+  </description>
+
```
theunknownxy:
```suggestion

  <categories>
​      <category>Game</category>
​      <category>Shooter</category>
  ​</categories>

```

flatpak/jp.yvt.OpenSpades.appdata.xml
```diff
@@ -0,0 +1,66 @@
...
+  <launchable type="desktop-id">jp.yvt.OpenSpades.desktop</launchable>
+
+  <url type="homepage">http://openspades.yvt.jp/</url>
+  <url type="bugtracker">https://github.com/yvt/openspades/issues</url>
+
```
theunknownxy:
```suggestion
  <url type="translate">https://crowdin.com/project/openspades</url>

```

flatpak/jp.yvt.OpenSpades.json
```diff
@@ -0,0 +1,119 @@
...
+          "sha256": "c9e6723aad729140d4af0771d780d20823ca5b0c2fc52cf5a9b2028e923d33d4"
+        }
+      ]
+    },
+    {
+      "name": "openspades",
+      "buildsystem": "cmake",
+      "build-options": {
+        "env": {
+          "CC": "clang",
```
theunknownxy: I almost forgot this: Is there are reason to use clang instead of gcc?

feikname:
> For example i don't know whether "social-chat" also includes "language-profanity". The game itself doesn't contain any "bad" words, but the language on public servers usually isn't that nice.

I will check this later today.

flatpak/jp.yvt.OpenSpades.json
kdeleteme: I removed the build options.

flatpak/jp.yvt.OpenSpades.json
```diff
@@ -0,0 +1,119 @@
...
+  "cleanup": [
+    "/include",
+    "/lib/pkgconfig"
+  ],
+  "modules": [
+    "shared-modules/glew/glew.json",
+    {
+      "name": "opus",
+      "sources": [
+        {
```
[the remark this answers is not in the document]
kdeleteme: I change it to pull the archive instead of from git.

feikname:
From what I understood, `language-profanity` should only be used if the game itself contains profanity, and is not applicable to chat: that's why there's a separate `social-chat` attribute.

So the resulting OARS should be:
```xml
  <content_rating type="oars-1.1">
    <content_attribute id="violence-cartoon">intense</content_attribute>
    <content_attribute id="violence-bloodshed">mild</content_attribute>
    <content_attribute id="social-chat">intense</content_attribute>
  </content_rating>
```

kdeleteme:
I edited the flatpak manifest to use the extra-data mechanism and added the OARS data. This looks good to me.

feikname:
Is the issue where configuration/player data is reset on every launch still present?

kdeleteme:
> Is the issue where configuration/player data is reset on every launch still present?

When I run it with flatpak-builder --run, it seems to save/load configuration data. But when I install and then run it, it resets the config.

feikname:
>When I run it with flatpak-builder --run, it seems to save/load configuration data. But when I install and then run it, it resets the config.

Just a guess, but could you please try try changing `--persist=.local/share/openspades/` to `--persist=~/.local/share/openspades/`?

If that doesn't work either we can do `--filesystem=home` for now and do better sandboxing later (along solving #686).

kdeleteme:
I tried your suggestions but they didn't work for me. I went with --filesystem=home

feikname:
So, all good to merge? 😃

kdeleteme:
Yup. Looks good to me. Finally :sweat_smile:

theunknownxy:
Okay I have no idea what flatpak is doing: when you persist `.local/share/openspades`, it stores the files at `$HOME/.var/app/jp.yvt.OpenSpades/.local/share/openspades/Resources/SPConfig.cfg`. But within the container the openspades directory ended up at `$HOME/.local/share/openspades/openspades/`.
And suddenly after playing a bit around with the flatpak-builder config, the directory now ends up at the correct location...

Can you try to persist `.local/share/openspades` (without trailing slash) and then before you run OS delete `$HOME/.var/app/jp.yvt.OpenSpades/.local/`? Maybe flatpak was confused, because the directory already existed?

kdeleteme:
Replaced `--filesystem=home` with `--persist=.local/share/openspades`

I tried it and it works for me.

feikname:
@yvt this is ready to merge.

PR #815, opened 2019-03-25
expand disconnect reasons

NotAFile:
add the new disconnect reasons

### DIFF
--- a/Sources/Client/NetClient.cpp
+++ b/Sources/Client/NetClient.cpp
@@ -752,14 +752,12 @@ namespace spades {
 		std::string NetClient::DisconnectReasonString(enet_uint32 num) {
 			switch (num) {
 				case 1: return _Tr("NetClient", "You are banned from this server.");
-				case 2:
-					// FIXME: this number seems to be used when per-IP connection limit was
-					// exceeded.
-					//        we need to check other usages
-					return _Tr("NetClient", "You were kicked from this server.");
+				case 2: return _Tr("NetClient", "You have too many connections to this server.");
 				case 3: return _Tr("NetClient", "Incompatible client protocol version.");
 				case 4: return _Tr("NetClient", "Server full");
+				case 4: return _Tr("NetClient", "Server was shut down");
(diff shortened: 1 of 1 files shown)


BR-:
Should we use [https://github.com/piqueserver/aosprotocol/issues/15](this idea) instead of adding more fixed strings?

NotAFile:
There's a few things with that:
 - Those messages aren't localized to the local language of the user
 - There are already installs out there using these new error codes. If we don't add them, they will show up as "Unknown Reasons" instead.
 - It really doesn't hurt to have more reasons here.

Sources/Client/NetClient.cpp
```diff
@@ -752,14 +752,12 @@ namespace spades {
 		std::string NetClient::DisconnectReasonString(enet_uint32 num) {
 			switch (num) {
 				case 1: return _Tr("NetClient", "You are banned from this server.");
-				case 2:
-					// FIXME: this number seems to be used when per-IP connection limit was
-					// exceeded.
-					//        we need to check other usages
-					return _Tr("NetClient", "You were kicked from this server.");
+				case 2: return _Tr("NetClient", "You have too many connections to this server.");
```
yvt: The word “you” might be inappropriate here because it’s not necessarily “you” who has connections (e.g., other people behind a NAT).

NotAFile: I had similar concerns but I was unsure how to phrase it instead. "Your IP" might be correct at the moment, but:
 - people, in general, are not that familiar with where their IP comes from, especially in larger networks and with the increased prevalence of CGNATs
 - this is a technical detail and might change in the future

Perhaps "Your location" or "Your network" is a good solution?

feikname: Seems like "Your network's public address has too many connections to this server" would be the best solution, even though some users may not understand it, I believe it's best to deliver the most accurate message.

Maybe a text below with additional info could be added including the common reasons this might be happening.

PR #825, opened 2019-05-04
detect end of map transfer with state packet

### LINKED ISSUE #826: Packet ID 60 breaks map loading
When piqueserver sends a ProtocolExtensionInfo packet with id 60 while the map is being loaded, I get stuck at the team selection screen and an error is printed to stdout.
`Map decoder returned error. Maybe we will get more data...`
Adding `reader.GetType() != 60 &&` after NetClient.cpp L611 fixes the problem.

NotAFile:
Previously, the end of map transfer was detected by a number of fallible
heuristics, including checking if the map is loadable on every non-map
packet, when a timeout had been reached, or the map size given by the
server was reached. This broke in a number of scenarios, including when
an unexpected packet was sent.

Instead, we now detect the end of the map transfer when we receive a
state packet, as intended by the protocol.

I have tested that this works for:
 - initial map load
 - map advances
 - map advances with pending reloads

Fixes #826

### DIFF
--- a/Sources/Client/ClientPlayer.cpp
+++ b/Sources/Client/ClientPlayer.cpp
@@ -315,8 +315,6 @@ namespace spades {
 
 			PlayerInput actualInput = player->GetInput();
 			WeaponInput actualWeapInput = player->GetWeaponInput();
-			Vector3 vel = player->GetVelocty();
-			vel.z = 0.f;
 			if (actualInput.sprint && player->IsAlive()) {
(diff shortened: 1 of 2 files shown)


NotAFile:
wait crap, wrong patches

NotAFile:
Should be good now. This PR also removes an unused variable I found. Adding it to this PR was totally on purpose.

yvt:
LGTM

PR #828, opened 2019-05-12
Add support for ExtInfo Packet

NotAFile:
as documented in https://github.com/piqueserver/aosprotocol/pull/22

### DIFF
--- a/Sources/Client/NetClient.cpp
+++ b/Sources/Client/NetClient.cpp
@@ -92,6 +92,7 @@ namespace spades {
 				PacketTypeHandShakeReturn = 32, // C2S
 				PacketTypeVersionGet = 33,      // S2C
 				PacketTypeVersionSend = 34,     // C2S
+				PacketTypeExtensionInfo = 60,
 
 			};
 
@@ -505,7 +506,7 @@ namespace spades {
 					auto &reader = readerOrNone.value();
 
 					try {
-						if (HandleHandshakePacket(reader)) {
+						if (HandleHandshakePackets(reader)) {
 							continue;
 						}
 					} catch (const std::exception &ex) {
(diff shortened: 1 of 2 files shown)


Sources/Client/NetClient.h
```diff
@@ -69,6 +76,10 @@ namespace spades {
 			std::shared_ptr<GameProperties> properties;
 
 			int protocolVersion;
+			std::unordered_map<uint8_t, uint8_t> extensions;
+			std::unordered_map<uint8_t, uint8_t> implementedExtensions {
```
yvt: Where is this `implementedExtensions` accessed from?

Sources/Client/NetClient.h
```diff
@@ -69,6 +76,10 @@ namespace spades {
 			std::shared_ptr<GameProperties> properties;
 
 			int protocolVersion;
+			std::unordered_map<uint8_t, uint8_t> extensions;
```
yvt: Add a short description.

Sources/Client/NetClient.cpp
```diff
@@ -714,6 +716,25 @@ namespace spades {
 			}
 		}
 
+		void NetClient::HandleExtensionPacket(spades::client::NetPacketReader& reader) {
+			std::unordered_map<uint8_t, uint8_t> implementedExtensions ( {{192, 1}, {120, 1}} );
```
yvt: Add `static const` because it's constant.

NotAFile: Sorry, it seems I messed up the patch once again. This wasn't supposed to be here.

NotAFile:
Should be good to go now, once the tests run through.

PR #830, opened 2019-05-19
Display a real progress bar when receiving a map

yvt:
This PR introduces a *real* progress bar displaying how much percentage of a map has been downloaded, instead of just the number of bytes downloaded, which didn't give players an idea how long it would take.

progressbar

`MapStart` is useless for this purpose because the server doesn't know the compressed size at the point when `MapStart` is sent due to progressive compression and just sends [a constant value](https://github.com/piqueserver/piqueserver/blob/05447fe32bacf7c5630654577140092d4f4e48be/pyspades/mapgenerator.py#L36-L41). The approach taken here is to start decoding the map data as it's being received and examine the amount of the decoded data. The correspondence between the decoded size and the compressed size is not linear, but gives players a better idea how longer the receiving process is going to take.


NotAFile:
How about showing the progress bar as percentage of columns sent? That's currently constant for all maps.

yvt:
> How about showing the progress bar as percentage of columns sent? That's currently constant for all maps.

Actually that’s what this PR does. I should’ve worded it better.

PR #833, opened 2019-05-24
Use model for block cursor instead of debug lines

otzade:
screenshots (updated):
inactive
active


NotAFile:
OpenSpade's render pipeline didn't support transparency, right?

I really like the dotted yellow one looks, I think the white one is a bit too much though. Perhaps both could use the same model, but different colors?

otzade:
Yeah the one with the corners only looks a lot better, but when using whe same model it's harder to see if the block is active or not. There isn't a way to set transparency to a model from what i see.

otzade:
It it were up to me I would remove the yellow cursor entirely and make it only visible when you can place a block

NotAFile:
btw, this reminds me of how Iceball used to have small voxel + floating in the air to make the boundary of the map visible. This would probably be horifically inefficient to do using the kv6 model code, but I do think it would be nice to have some marker for the map border.

otzade:
Iceball was the inspiration of this, it has a similar model cursor

yvt:
I don’t like cursors shaded like real in-game objects. Maybe try an emissive material (for which I added support in [a recent commit](https://github.com/yvt/openspades/blob/51af48e98bb1581ff56f295f97395fece2efcaef/Sources/Core/VoxelModelLoader.h#L35-L54)) and [`castShadow = false`](https://github.com/yvt/openspades/blob/51af48e98bb1581ff56f295f97395fece2efcaef/Sources/Client/IRenderer.h#L49)?

otzade:
Btw color correction gives the white cursor a slight red tint even on neutral backgrounds

yvt:
> Btw color correction gives the white cursor a slight red tint even on neutral backgrounds

The color correction compensates for overly saturated sky colors by multiplying an opposite color (hence the red tint). Not much a problem IMO.

yvt:
Ready to merge?

NotAFile:
Just tried it out, I really like this! The double border for neighboring blocks looks a tiny bit weird but it's easy to get used to!

Issue #834, opened 2019-05-24
Dynamic lights not working


otzade:
Dynamic lights don't work anymore, just compiled 0.1.2 and they work there

NotAFile:
what exactly do you mean with dynamic lights? Both muzzle flashes and flashlights have worked fine for me.

(Also, would probably be really helpful if you could do a `git bisect` on this)

yvt:
Here's the list of possibly relevant commits: https://github.com/yvt/openspades/compare/e1abcf9c41982445bd15f8dce21149d2da3254a8...ac6bc0161751d9b3e2e04d149f4453cfa4fee520

otzade:
Muzzle and grenade flashes
bisect:
91ebbdb4967beca8da9eda8367be658b9c96ed62 `is the first bad commit`

otzade:
Not sure what artifact this commit is talking about but on my amd card it actually added an artifact.
91ebbdb4967beca8da9eda8367be658b9c96ed62 not applied:
![before commit](http://i63.tinypic.com/4j47sx.jpg)
91ebbdb4967beca8da9eda8367be658b9c96ed62 applied:
![after commit](http://i64.tinypic.com/2vb4hna.jpg)

note that the artifact only appears on my amd gpu, not the intel one
also after this commit dynamic lights don't appear on both gpus

PR #835, opened 2019-05-25
Fix issues with dynamic lights

yvt:
This PR fixes the two issues mentioned in #834.


yvt:
@otzade Can you see if this fixes the issue?

otzade:
Fixed the artifacts and the dlights :^)
tried on both gpus

Issue #780, opened 2018-10-13
Abort on assertion failure


yvt:
It should result in a program termination (preferably with a core dump and helpful message) rather than throwing an exception for various reasons.

theunknownxy:
Here are my thoughts about the network data validation issue: Receiving malformed data is not an logic error. For example an malicious server shouldn't be able to crash the client by triggering an assertion. The client should rather disconnect from the server, which means that normal error handling is more appropriate (using exceptions or error codes).

NotAFile:
Network code is especially critical in that regard, since special strictness on some things e.g. out of range packet IDs and values might make expansion of the protocol harder.

theunknownxy:
@NotAFile I don't agree, because it's not possible to distinguish between an broken packet and a recently defined packet. The client wouldn't know if the unknown or corrupt data is important. The server just shouldn't send new packets to an "old" client.

NotAFile:
You're kind of right about that now, as OpenSpades now implements packet 60 extinfo that can be used to query capabilities of the client. But keep in mind that things such as the handshake, version and extinfo packets are only possible because voxlap AoS ignores unknown packets. Implementing packet 60 extinfo in the server was also made more difficult because OpenSpades did not really ignore it, which then required a hardcoded workaround for OS <=0.1.3 on the server side.

PR #836, opened 2019-05-26
Little code changes proposal.

devnexen:
Modern C++ stricter with member initialisation order.
Destructor possibly throwing exception through assertions.
In debug mode, pretty important memory leak with zip archives.

### DIFF
--- a/Sources/Core/IStream.cpp
+++ b/Sources/Core/IStream.cpp
@@ -26,7 +26,7 @@
 #include "IStream.h"
 
 namespace spades {
-	IStream::~IStream() {}
+	IStream::~IStream() throw(Exception) {}
 	int IStream::ReadByte() { SPUnsupported(); }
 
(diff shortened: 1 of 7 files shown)


Sources/Core/IStream.cpp
```diff
@@ -26,7 +26,7 @@
 #include "IStream.h"
 
 namespace spades {
-	IStream::~IStream() {}
+	IStream::~IStream() throw(Exception) {}
```
theunknownxy: I don't think using exceptions specifications is a good idea. They were deprecated in C++11 and removed in C++17: https://en.cppreference.com/w/cpp/language/except_spec

yvt: I guess I missed that part. As a related matter, there’s #780 regarding deprecating exceptions thrown by assertion failures.

Issue #841, opened 2019-07-05
Hot skin reloading


yvt:
As every modder of OpenSpades knows, restarting OpenSpades every time you make a small modification is quite a hassle. This issue proposes a hot code reloading mechanism for skin scripts.

## Task List

- Support unloading models loaded via `IRenderer::RegisterModel`.
- Support unloading audio chunks loaded via `IAudioDevice::RegisterSound`.
- Load `Scripts/Skin/*` as a separate AngelScript module.
- Make `ScriptFunction` accept a module via a parameter.
- Stop storing `ScriptFunction` in `static` variables in `ClientPlayer`.
- Add an internal interface to recreate relevant AngelScript objects when the module is reloaded.
- Add UI for triggering a reload.

## Requirements

- The implementation should properly unload the old modules and release associated resources. All references to the old modules must be destroyed.
- Detectable compilation/runtime errors should not be a hard error (should not terminate the application like they do now).
    - However, compilation errors detected during the first-time load should be a hard error and reported as soon as possible, following the fail-fast principle.

Issue #57, opened 2013-09-08
Replay packets.


dany-on-demand:
1. Save packets recieved (and sent) in order to a .demo file (I hacked this together, but it saves in to an uncompressed ascii file, a compressed binary file would be better, also I don't feel what I wrote complies with the way it's supposed to be used)
2. If you type demo name/drop a demo file into the launcher, a game would launch, connect to a "fake server" that would stream packets from the file.

Uses:
1. Admins can recieve demos from users and see if players complied with the rules long after they've disconnected. Less cheaters!
2. No more need for practically supplementary wallhack programs that really shouldn't be in the public.
3. Replays of matches for noobs to learn how the pros play :)
4. At some point, we might add some kind of super-simple cutting and advanced camera controls, this way we could make much nicer trailers, and even machinima! Would be neat as fuck to make propaganda videos rather than posters! :D

noway421:
@dany-on-demand pull request? looks awesome.

VelocityRa:
Well, this *is* from 3 years ago...

PR #842, opened 2019-07-13
Add internal console window

yvt:
This PR adds an internal console window similar to the one seen in certain old-school PC games.

Screen Shot 2019-07-14 at 20 42 10 copy

Screen Shot 2019-07-15 at 0 24 51 copy

- The console window serves as an interface to hidden functionalities such as hot skin reloading (#841), [in-game config variable modification](https://github.com/yvt/openspades/wiki/Config-Variables), and demo recording (#57).
- It can be accessed anywhere in the game including `MainScreen`, since it's implemented in a lower level (think of Windows NT's <kbd>Ctrl</kbd>+<kbd>Alt</kbd>+<kbd>Delete</kbd>).
    - Technically, it can be made invokable even from `StartupScreen`, but I decided not to do so.
- In addition to the existing destinations, `SPLog` is also redirected to this new console window. The history size is limited and old entries are removed from memory as new one arrives (this is mandatory because the amount of log produced by `SPLog` is large).
- It's toggled by <kbd>\`</kbd>. <kbd>F1</kbd> is also accepted because some keyboard layouts don't have <kbd>\`</kbd>.
- It's modal; it intercepts all inputs when active.
- Unlike some implementations, it is **not** dual-purposed as a chat window.

## Task List

- [x] Add `View::RunFrameLate`
- [x] Add `ConsoleScreen` (a `View` that wraps another `View` to insert the console window 2D layer and intercept the input)
- [x] Implement the user interface (`ConsoleWindow`)
- [x] Deprecate the old in-game config variable modification functionality that uses the chat window. Show an alert whenever this feature is used to encourage the transition.
- [x] Add something like `View::RunCommand`
- [x] "Launch title"
    - [x] `varname value` — In-game config variable modification
    - [x] `savemap`
    - [x] (Maybe?) `cleargfxcache` — Unloads voxel models, forcing reload
    - [x] (Maybe?) `clearsfxcache` — Unloads audio chunks, forcing reload
    - [x] `help`
- [x] Autocompletion
- [x] Redirect `SPLog`


yvt:
All planned features are done - will merge in 24 hours unless any critical issues are found.

Issue #847, opened 2019-08-17
Stuck at loading map change on 0.76 server.


Tuupertunut:
Openspades doesn't load new maps on map change on 0.76 servers. It gets stuck on the screen with "Loading snapshot".

**Scope of the problem**

Occurs on all Openspades versions since at least 0.1.0 as well as on the latest master.
Doesn't occur on the original Ace of Spades 0.76 client.

Tested using PySnip server 0.76 version fd74ccc.

**How to reproduce**

- Download and run latest PySnip 0.76 release from https://github.com/NateShoffner/PySnip/releases
- Download and run latest Openspades
- Join the server
- Login as admin with `/login adminpass1`
- Advance map with `/advance`

You should see Openspades stuck at loading screen.

Clogal:
Here's something:
https://github.com/BR-/openspades/commit/d4a7934291373c97bfbc6774729a96b14edf343e

Tuupertunut:
I can confirm that commit does fix it.

Tuupertunut:
I made a PR of that commit.

PR #848, opened 2019-09-22
Fixed map loading with protocol 0.76.

Tuupertunut:
This solution is copied from a commit in another fork:
https://github.com/BR-/openspades/commit/d4a7934291373c97bfbc6774729a96b14edf343e

I don't really understand what the commit does, but it fixes the issue with map loading on 0.76 servers (#847).


NotAFile:
Right, basically what's happening is that 0.76 servers need to hear that we don't have the map cached before it will send the map.

Conticop:
I just tested this, and the client still errors about incompatibility with this change when trying to join v0.76 server.

Tuupertunut:
@Conticop That sound like a separate issue. This PR fixes map loading when a 0.76 server changes map (https://github.com/yvt/openspades/issues/847), not when joining the server.

PR #858, opened 2019-10-07
Fix aim-down-sight sound when ammo count is zero

half-cambodian-hacker-man:
'IsReadyToUseTool' checks ammo count, whereas this method does all the same checks. (At least, for a client-sided ToolWeapon.)

Fixes #858.


yvt:
Thanks!

PR #861, opened 2019-10-15
Fix dryfire

### LINKED ISSUE #853: DryFire sound does not work
It seems that it worked properly before this commit : https://github.com/yvt/openspades/commit/2046afd660a6907787a9b74ef3b9270bccbe4c04

abmyii:
Fixes #853. I'm unsure if there should be a timer on dry-fire too, but I've disabled it for now.
Thanks for this great game @yvt!

### DIFF
--- a/Sources/Client/Client_Update.cpp
+++ b/Sources/Client/Client_Update.cpp
@@ -426,11 +426,6 @@ namespace spades {
 					winp.primary = false;
 					winp.secondary = false;
 				}
-
-				// stop firing if the player is out of ammo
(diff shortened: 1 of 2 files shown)


abmyii:
Any updates?

abmyii:
Ping.

Clogal:
In voxlap dry-fire sound also gets played when sprinting or changing tool, and it doesn't sound like if it has a timer

abmyii:
@yvt Will this not be merged?

PR #871, opened 2019-11-27
Fix use after free in Main.cpp (issue #846)

### LINKED ISSUE #846: Use after free in Main.cpp
On latest clang trunk, the code
```
const char *text = ("Content of this directory moved to " +
							                    xdg_data_home + "/openspades")
							                     .c_str();
```
in https://github.com/yvt/openspades/blob/master/Sources/Gui/Main.cpp#L428

causes the warning 
```
warning: object backing the pointer will be destroyed at the end of the full-expression [-Wdangling]
```
because `xdg_data_home  + ...` creates a temporary `std::string` which is destroyed at the end of the line, and `text` then points into that destroyed string.

dtomlinson-ga:
Create c-string to store filepath before concatenating with output message. [Fix #846](https://github.com/yvt/openspades/issues/846)

### DIFF
--- a/Sources/Gui/Main.cpp
+++ b/Sources/Gui/Main.cpp
@@ -425,9 +425,8 @@ int main(int argc, char **argv) {
 						SDL_RWops *io = SDL_RWFromFile(
 						  (home + "/.openspades/CONTENT_MOVED_TO_NEW_DIR").c_str(), "wb");
 						if (io != NULL) {
-							const char *directory = (xdg_data_home + "/openspades").c_str();
-							const char *text = ("Content of this directory moved to " + directory).c_str();
-							io->write(io, text, strlen(text), 1);
+							std::string text = ("Content of this directory moved to " + directory);
+							io->write(io, text.c_str(), text.length(), 1);
 							io->close(io);
 						}
 					}
--- a/Sources/Gui/Main.cpp
+++ b/Sources/Gui/Main.cpp
@@ -425,7 +425,7 @@ int main(int argc, char **argv) {
 						SDL_RWops *io = SDL_RWFromFile(
 						  (home + "/.openspades/CONTENT_MOVED_TO_NEW_DIR").c_str(), "wb");
 						if (io != NULL) {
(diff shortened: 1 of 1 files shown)


dtomlinson-ga:
I don't know if this is the best way to go about doing this, but I wanted to dip my toes into contributing to this project. Let me know if anything needs to change :)

Sources/Gui/Main.cpp
```diff
@@ -425,9 +425,8 @@ int main(int argc, char **argv) {
 						SDL_RWops *io = SDL_RWFromFile(
 						  (home + "/.openspades/CONTENT_MOVED_TO_NEW_DIR").c_str(), "wb");
 						if (io != NULL) {
-							const char *text = ("Content of this directory moved to " +
-							                    xdg_data_home + "/openspades")
-							                     .c_str();
+							const char *directory = (xdg_data_home + "/openspades").c_str();
+							const char *text = ("Content of this directory moved to " + directory).c_str();
```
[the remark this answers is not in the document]
dtomlinson-ga: Thank you for the helpful feedback! I wasn't familiar with this usage of the `const char*` declaration, but looking around the code, declaring as a `std::string` as you described is definitely more prevalent.

Sources/Gui/Main.cpp
```diff
@@ -425,10 +425,8 @@ int main(int argc, char **argv) {
 						SDL_RWops *io = SDL_RWFromFile(
 						  (home + "/.openspades/CONTENT_MOVED_TO_NEW_DIR").c_str(), "wb");
 						if (io != NULL) {
-							const char *text = ("Content of this directory moved to " +
-							                    xdg_data_home + "/openspades")
-							                     .c_str();
-							io->write(io, text, strlen(text), 1);
+							std::string text = ("Content of this directory moved to " + directory);
```
yvt: Build is failing because `directory` is not defined.

dtomlinson-ga: ... I need to stop writing these at 1 am. Sorry for the rookie mistake, should be fixed now!

PR #874, opened 2019-12-04
packaging: add Gnu Guix pkg for OpenSpades

o-nly:
to build with guix:
`guix package -f guix/guix.scm'

[Gnu Guix](https://guix.gnu.org/) is a Free/Libre package manager. I was playing OpenSpades yesterday on a different OS. I thought I'd try to package it for Guix today.

give it a spin and report problems.

### DIFF
--- a/guix/guix-pkg-1.patch
+++ b/guix/guix-pkg-1.patch
@@ -1,0 +1,34 @@
+From e28fbb02aa0c334e2382f5996417ee846faba7c5 Mon Sep 17 00:00:00 2001
+From: Amar Singh <nly@disroot.org>
+Date: Wed, 4 Dec 2019 21:29:43 +0530
+Subject: [PATCH] guix-patch: do not download nonfree content.
+
+---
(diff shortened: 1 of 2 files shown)


godwhoa:
Another way to install openspades is cool and all, but are you willing to maintain this in the future?

o-nly:
yeah, i do.

NotAFile:
I don't think it's a great idea to maintain packaing for every distribution and in-tree, especially with the unmaintained ones present there already.

I think maintaining this packaging info in a separate repo like upstream Guix does would be much preferrable.

o-nly:
> I think maintaining this packaging info in a separate repo like upstream Guix does would be much preferrable.

Upstream is strict about FOSS licensing. Non-strict [Nonguix](https://gitlab.com/nonguix/nonguix) will be a suitable home for this package.

NotAFile:
There had been work on freely licensed assets, but I think that died down due to a lack of interest.

PR #877, opened 2020-01-11
AngelScript: fix build on DragonFly BSD, NetBSD and sparc64

goleo108:


goleo108:
there should be more elegant patch for as_memory.cpp

NotAFile:
Maybe you should send this patch upstream to https://sourceforge.net/projects/angelscript/?

Or perhaps this has already been fixed upstream, in which case the vendored version should be updated.

goleo108:
@NotAFile well, the problem is that in my experience projects using angelscript avoid updating it because its semantic versioning is wrong, API itself breaks on every update, as a former maintainer of OpenBSD ports and Ravenports I tried to follow the policy of building against system-wide libraries, however, in angelscript case I had only 3 options sorted by difficulty:
1. build internal angelscript and statically link it
2. port several versions of angelscript and patch build systems of dependants to use the different naming of libraries
3. patch dependants to always use latest angelscript available

angelscript is not so popular to choose option 2, so the only adequate choices are option 1 and option 3, but with my current skills I can only go with option 1

PR #884, opened 2020-03-01
Memory overlap fix proposal.

devnexen:


NotAFile:
How about fixing this in upstream AngelScript instead of here?

PR #889, opened 2020-04-15
Removed unnecessary call to GetHorizontalLength() in Player.cpp

hourai-dev:
Now the result of GetHorizontalLength() is just stored in a variable and that variable is used for the checks, instead of unnecessarily calling the function twice.


hourai-dev:
Thanks for reviewing my PR, I have done the corrections now.

yvt:
Everything looks good now. Thanks!

PR #894, opened 2020-04-28
Added a selection menu for the next spawn loadout

ptrstr:
This adds a new button under "Spawn" in the Limbo menu to set the next spawn's loadout. This allows you to set your next loadout for when you'll die rather than killing yourself if you want another gun and not continuing the battle you're currently in. List of issues with the current implementation:

- Might be excessive to check if player is alive on every LocalPlayer update
- You can see the button when you just joined a game. I disabled it for when you're in spectator mode since you can't die but couldn't figure out how for the initial menu.

### DIFF
--- a/Sources/Client/Client.cpp
+++ b/Sources/Client/Client.cpp
@@ -519,6 +519,18 @@ namespace spades {
 			}
 		}
 
+		void Client::NextSpawnPressed() {
+			WeaponType weap = limbo->GetSelectedWeapon();
+			int team = limbo->GetSelectedTeam();
+			inGameLimbo = false;
+			if (team == 2)
+				team = 255;
+
+			this->hasNextSpawnConfig = true;
+			this->nextTeam = team;
+			this->nextWeapon = weap;
+		}
+
 		void Client::ShowAlert(const std::string &contents, AlertType type) {
 			float timeout;
 			switch (type) {
--- a/Sources/Client/Client.h
+++ b/Sources/Client/Client.h
@@ -286,10 +286,15 @@ namespace spades {
 
 			bool inGameLimbo;
 
+			bool hasNextSpawnConfig;
+			int nextTeam;
+			WeaponType nextWeapon;
+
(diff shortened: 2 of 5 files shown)


Sources/Client/Client.cpp
```diff
@@ -519,6 +519,18 @@ namespace spades {
 			}
 		}
```
yvt, line 520: Reset `hasNextSpawnConfig` in `SpawnPressed`. Consider the following scenario:
1. While being alive, the player sets the next spawn configuration C1. `hasNextSpawnConfig` ← `true`
2. The current round concludes, and the game switches to the next map.
3. The player decides an initial spawn configuration C2 and spawns.
4. The player dies.
5. The player spawns with C1 because `hasNextSpawnConfig == true`.

This behavior would be surprising.

ptrstr:
Added all your suggestions. Also, thank you for the scenario idea. Didn’t think of it. Not sure I implemented the `stmp::optional` correctly though since either way it’d work.

ptrstr:
Thanks for the help. Really nice of you to have made this cheatsheet 👍. Do more C than C++ that's why I was so alienated by those `std::` and `stmp::` types. Fixed it and tested.

ptrstr:
Added support for spectator mode 👍

yvt (approved):

VierEck:
> * Might be excessive to check if player is alive on every LocalPlayer update

u can do the check in this function instead which is only called when a killaction packet is received
https://github.com/yvt/openspades/blob/2d91a3f7fa4aff02767c8b2c7b59c32acf5a733f/Sources/Client/Client_Update.cpp#L783-L784
and send the next spawn loadout if victim == localplayer
https://github.com/yvt/openspades/blob/2d91a3f7fa4aff02767c8b2c7b59c32acf5a733f/Sources/Client/Client_Update.cpp#L812-L819


it would have been better to have made an additional dedicated feature branch for this and made the pull request from there instead of master branch.

PR #905, opened 2020-08-03
Pubovl (including extras)

### LINKED ISSUE #807: add pubovl spectating
Currently pubovl-like tools are the only ways of recording the use of hacks in game for evidence for admin. There is a modified version of 0.0.12 that has pubovl added but this feature is needed in future releases until something better can be developed.

http://aloha.pk/index.php?topic=8558.0

Conticop:
Fixes #807
Original author: dd

I took some time to get [*Pubovl*](http://aloha.pk/index.php?topic=8558.0) compatible with the latest `master`, and it has arrived here now.
I have tested it ingame, all works fine.
(See the above forum post for more info about *Pubovl* and its features.)

----

> The bigger issue for me is that it makes it much easier to write a wallhack. All you need to do is delete the "only allow wallhack if you are spectator" check.

I thought, well it was made available to the public since a long time ago, with binaries precompiled and available for download from the same forum post above. So what the heck.
There'll always be cheaters in such game(s). Merge this to make admins' lives easier.


Conticop:
I suppose I should have removed the comments...

Conticop:
There, this is ready for review.
cc @yvt

Sources/Draw/GLOptimizedVoxelModel.cpp
```diff
@@ -613,9 +620,9 @@ namespace spades {
 			device->ActiveTexture(0);
 			device->BindTexture(IGLDevice::Texture2D, 0);
 		}
-
+
```
Conticop, line 623: How was this born

Resources/Shaders/BasicBlockOutlines.program
```diff
@@ -0,0 +1,4 @@
+Shaders/BasicBlockOutlines.fs
+Shaders/BasicBlockOutlines.vs
+*shadow*
+Shaders/Fog.vs
```
feikname, line 4: Missing end of line.

Clogal:
The extras should be removed (cg_outlines, cg_textures), as they give some visibility advantage.
The player model already have neon-like colors, but when raising texture strength to more than 20 it gives extra constrast, that and cartoon-like outlines would give the perfect visibility.

shot0001
shot0002

Haxk20:
@Conticop Could you get this updated with the latest master changes ?
@yvt A little ping
This is an absolutely amazing change. While yes cheaters could easily enable ESP in real game.
Tho i would argue that it is FAR far easier for a cheater to download already made hacks then poking in the code.
And it is not that hard to add some multiple layers of checks in the code if they wanted to make ESP accesible to real gameplay.

JimPix1:
@yvt Ping

Sources/Draw/GLRenderer.cpp
```diff
@@ -846,18 +914,117 @@ namespace spades {
...
 
 			device->FrontFace(IGLDevice::CW);
 
-			{
+			bool visiblePlayers[32];
+
+			if (!client::Client::WallhackActive()) {
 				GLProfiler::Context p(*profiler, "Non-mirrored Objects");
-				RenderObjects();
+				RenderObjects(false);
```
yvt, line 927: Don't refer to `Client` directly from the renderer layer. Introduce adequate abstractions in the renderer API.

yvt (changes requested):
Sorry for the delay. Although it's marked as draft now, I'm going to submit a review anyway to explain why it cannot be merged in the current state:

The largest problem is already stated by @siecvi: In addition to fixing #807, the PR introduces multiple features irrelevant to the primary change, such as the outline ([NotTodo](https://github.com/yvt/openspades/wiki/NotTodo) 2-3) and texturing effects and even global sound volume control. The PR must be kept minimal and focus on a single logical change.

Another problem concerns code quality - bypassing abstraction, the use of global variables, and such. The existing code is bad as it is, but let's not make it even worse.

JimPix1:
Thanks for taking the time to reply, yvt. Hoping these problems get fixed 🤞

PR #908, opened 2020-09-20
Content class little code simplification.

devnexen:


yvt:
Couldn't it be simply removed? The compiler automatically [defines](https://en.cppreference.com/w/cpp/language/copy_constructor#Implicitly-declared_copy_constructor) one when a copy constructor is not defined nor deleted.

yvt:
Thanks!

PR #913, opened 2020-10-08
.

DeletedUser92:
.


NotAFile:
I'm not familiar with rich presence, but should this be made optional at compile time? It appears to be a proprietary library.

PR #924, opened 2020-12-13
Adopt vcpkg and streamline the build process

### LINKED ISSUE #739: Consider using Vcpkg for managing the external dependencies on Windows and macOS builds
https://github.com/Microsoft/vcpkg

**Pros:**

- Developers no longer have to trust the pre-compiled package maintainer (= me!) not to put malicious things in the pre-compiled opaque binary blob
- Developers (including me) will automatically have an access to the latest version of dependent libraries
    - Pre-compiled libraries are built manually and aren't updated for quite a long time.

**Cons:**

- Developers have to trust Microsoft and the ecosystem
- Extra step for installing vcpkg

### LINKED ISSUE #802: Japanese IME and Korean Input do not work on Windows
Tested on Windows 8.1 and in a Windows 10 VM. Strangely enough, (Simplified) Chinese IME works.

I am not quite familiar enough with Korean input to say what should generally happen, but none of the input makes it through.

Expected functionality for JP IME would be:
Type the romanized text, and while you're typing, the text changes to its hiragana equivalent (or otherwise depending on IME mode) in real time, appearing with an underline beneath it. Arrow keys Up/Down or [Shift+]Tab allow you to choose suggestions for the whole of the text, or if you press space, you can change individual parts of the text into the IME's suggestions via Left/Right, and Up/Down or [Shift+]Tab. After you hit Enter, the underlined text is inputted into the field and the underline goes away.

What actually happens is... nothing. The IME doesn't work, the game just eats all of the input, even the ASCII characters. Typing as though the IME were working doesn't work.

Seeing as the IME doesn't work, I'm unsure as to whether or not there are UI elements displaying the IME's suggestions. If not, that's another thing to look into.

Typing the text in another program and pasting it into OpenSpades does _work_, but this is obviously not ideal.

yvt:
This PR makes the following changes to OpenSpades' standard build process for macOS and Windows.

- It now uses [vcpkg](https://github.com/Microsoft/vcpkg) to fetch and build external dependencies. This makes their builds fully reproducible and reviewable and obviates the need to trust the prebuilt library packages' maintainer.
    - Closes #739.
    - vcpkg brings the latest versions of the libraries, which closes #802.
- The CMake build process for macOS now produces a complete application bundle, which is almost ready for distribution.
- The macOS build process now automatically downloads required assets.
- The external dependencies are now built as static libraries. This means the following benefits:
    - Their `.dll` files no longer clutter the Windows binary distribution.
    - Static linking decreases the total code size.
- The Xcode project has been removed because the CMake build process can do the job.
- The new CI pipeline runs in GitHub Actions.

## Tasks

- [x] Get the Windows build working
- [x] Get the macOS build working
- [x] CI: Windows builds
- [x] CI: macOS builds
- [x] Document the revised build process for Windows
- [x] Document the revised build process for macOS
- [x] Get rid of the `Externals` directory
- [ ] Update [the publishing process](https://github.com/yvt/openspades/wiki/Publishing-a-New-Release)


yvt:
Static linking in action (left: the latest release, right: this PR)

Screenshot from 2020-12-18 02-25-25

PR #935, opened 2021-01-08
Sharpening post-processing effect

yvt:
This effect kills two blocky birds with one stone. Firstly, it enhances the rendered image's clarity for artistic reasons. Secondly, it somewhat offsets OpenSpades' fog density, which is often claimed to be denser than the original client. Technically, the fog density function is mostly identical between these two clients. However, OpenSpades applies the fog color in the linear color space, which is physically accurate but has a tendency to strengthen the effect because the human light perception is logarithmic.

This PR's sharpening effect enhances local color differences to make distant objects look as clearly as they would do in the original client, on average. The effect intensity needs to be determined carefully not to over-enhance the image. The mathematical basis for the correct intensity determination is documented in the source code.

The sharpening effect is incorporated into the existing color correction pass. The effect amount can be adjusted by `r_sharpen` config variable.


yvt:
# Screenshots

2021-01-17 TAA and Sharpening


<details>
<summary>FXAA</summary>

none

</details>
<details>
<summary>FXAA + Sharpening</summary>

sharp

</details>
<details>
<summary>FXAA + Temporal AA</summary>

taa

</details>
<details>
<summary>FXAA + Temporal AA + Sharpening</summary>

taa+sharp

</details>
<details>
<summary>Vanilla Client</summary>

Screenshot from 2021-01-17 13-47-08 copy

</details>

Issue #409, opened 2016-07-14
Volumetric fog: rescale the fog density function to match the original one


yvt:

HuskyGFX:
I am curious: Ace Of Spades was never open-source so how do you find out details like this? Just hacking the original client somehow?

10se1ucgo:
mat^2 reverse engineered the 0.60 protocol a few years ago and created pyspades, the open source Ace of Spades server implementation. That's basically the only source for documentation for the protocol.

noway421:
In one of the antihack scripts (written by Mat IIRC) there was fog distance (range of visibility) which was apparently found by an experiment. That still doesn't tell the the visibility relationship to distance, but it might be just linear given the simplicity of voxlap renderer.

noway421:
>https://github.com/matpow2/pyspades-userscripts/blob/master/aimbot2.py#L86

if not linear then it is most probably quadratic

yvt:
Volumetric Fog Level 2 introduced by #941 resolves this issue. The old implementation will be dropped at some point.

PR #941, opened 2021-01-29
Volumetric Fog Level 2

yvt:
This PR adds a new volumetric fog implementation, which can be selected by `r_fogShadow 2`.

- The new implementation calculates the fog density in the same way as the original client does (#409).
- The influence factors of shadow casters are weighted according to their distances to the camera. Shadows don't "pop up" like they do in the Level 1 implementation.
- It's significantly more demanding compared to the Level 1 implementation but not as much as current-gen games.
- It supports global illumination. (The non-GI path is unimplemented, so it'll fall back to Level 1 if GI is disabled.)
- Being implemented by naive ray marching, its code is much simpler compared to the Level 1 one.

# Screenshots

shot0175

shot0189

shot0193


yvt:
# Screenshots

Demonstrating soft shadowing and a more physically-correct ambient term:

shot0254
shot0256


<details>
<summary>Level 1 (for comparison)</summary>

shot0255
shot0257


</details>

yvt:
shot0311
shot0290

Jipok:
My image is very uncomfortable - everything is too gray. Is this a normal rendering?
<details>
  <summary>Settings</summary>

image

</details>
<details>
  <summary>Level 1</summary>

image

</details>
<details>
  <summary>Level 2</summary>

image

</details>

yvt:
It's rendered as intended. The current implementation intensifies the effect in a close range. This was implemented in commit f6912b7a1da2fb68c6b7edb1565133f8fbb04c10 to accentuate the light shaft effect and justify the extreme processing cost incurred by the effect. Try reverting this commit and tell me what you think.

Jipok:
<details>
  <summary>Commit ga517f5ae</summary>

image
image
image

</details>
Lots of strange artifacts such as random glow or loss of clarity or black dots. Also, the fog (sky) should not be black.

But in some places the visibility is much better than on the latest version.

yvt:
I think you have reverted more than just commit f6912b7a1da2fb68c6b7edb1565133f8fbb04c10.

yvt:
Reverted the close-range fog density correction in commit 70443fe. I've been playing with both versions for a while and, turns out, don't like it.

Jipok:
Before commit:
image
After commit:
image
Got better. Less gray.

Jipok:
For myself, I prefer to play without global illumination. Best Visibility: 
image

Jipok:
I may have chosen the wrong place to complain. But I just can't figure out why these options make everything gray at all.

yvt:
Global illumination changes the luminance of shadows, which inevitably affects the outcome of auto-exposure and all post-effects that follow. With so many rendering steps involved, it's hard to say how exactly this affects the final rendered image.

yvt:
shot0526

shot0538

shot0534

PR #945, opened 2021-03-07
Fix crash due to shader program linking error

### LINKED ISSUE #944: OpenSpades does not even want to launch
Ok so, i get this when i try to launch `Vertex info
-----------
0(42) : error C3002: call to undefined function "void PrepareForRadiosityForMap_Map(vec3, vec3, vec3);"

at /home/maksymilian/openspades/Sources/Draw/GLProgram.cpp:67
void spades::draw::GLProgram::Link() at GLProgram.cpp:51
spades::draw::GLProgram* spades::draw::GLProgramManager::CreateProgram(const string&) at GLProgramManager.cpp:85
spades::draw::GLProgram* spades::draw::GLProgramManager::RegisterProgram(const string&) at GLProgramManager.cpp:57
void spades::gui::MainScreen::DoInit() at MainScreen.cpp:299
virtual void spades::gui::MainScreen::RunFrame(float) at MainScreen.cpp:232
virtual void spades::gui::ConsoleScreen::RunFrame(float) at ConsoleScreen.cpp:182
void spades::gui::SDLRunner::Run(int, int) at SDLRunner.cpp:427
void spades::gui::Runner::Run() at Runner.cpp:70
void spades::gui::Runner::RunProtected() at Runner.cpp:42
int main(int, char**) at Main.cpp:317
`

hourai-dev:
When built from master, Openspades would crash at startup with the
following error:
```
Error while linking a program: 'Shaders/BasicBlock.program'

error: unresolved reference to function 'PrepareForRadiosityForMap_Map'
```
The culprit was said function's definition in `MapRadiosityNull.vs`
which was missing a parameter. Fixes #944

### DIFF
--- a/Resources/Shaders/Shadow/MapRadiosityNull.vs
+++ b/Resources/Shaders/Shadow/MapRadiosityNull.vs
@@ -24,7 +24,7 @@ void PrepareForRadiosity_Map(vec3 vertexCoord, vec3 normal) {
 	hemisphereLighting = 1. - normal.z * .2;
 }
 
-void PrepareForRadiosityForMap_Map(vec3 vertexCoord, vec3 normal) {
+void PrepareForRadiosityForMap_Map(vec3 vertexCoord, vec3 centerCoord, vec3 normal) {
 	hemisphereLighting = 1. - normal.z * .2;
 }
 


yvt:
Thanks!

PR #952, opened 2021-05-05
Add volume adjustment for chat beeps

Pablo-Dallegri:


Sources/Client/Client.cpp
```diff
@@ -55,7 +55,7 @@
 
 #include "NetClient.h"
 
-DEFINE_SPADES_SETTING(cg_chatBeep, "1");
+DEFINE_SPADES_SETTING(cg_chatBeep, "100");
```
yvt, line 58: The existing users will keep using the old value of `cg_chatBeep`, so changing the value range or the default value is not recommended. I suggest using a 0–1 range (with intermediate values).

Sources/Client/Client.cpp
```diff
@@ -667,9 +667,11 @@ namespace spades {
 				NetLog("[Team] %s (%s): %s", p.GetName().c_str(),
 				       world->GetTeam(p.GetTeamId()).name.c_str(), msg.c_str());
 
-			if ((!IsMuted()) && (int)cg_chatBeep) {
+			if (!IsMuted()) {
 				Handle<IAudioChunk> chunk = audioDevice->RegisterSound("Sounds/Feedback/Chat.opus");
-				audioDevice->PlayLocal(chunk.GetPointerOrNull(), AudioParam());
+				AudioParam params;
+				params.volume = (int)cg_chatBeep / 100.f;
```
yvt, line 673: You can cast it to `float`.

Resources/Scripts/Gui/Preferences.as
```diff
@@ -645,7 +645,8 @@ namespace spades {
                                     array<int> = {2, 1, 0});
 
             layouter.AddHeading(_Tr("Preferences", "Feedbacks"));
-            layouter.AddToggleField(_Tr("Preferences", "Chat Notify Sounds"), "cg_chatBeep");
+            layouter.AddSliderField(_Tr("Preferences", "Chat Notify Sounds"), "cg_chatBeep",
+                                    0, 100, 1, ConfigNumberFormatter(0, "%"));
```
yvt, line 649: Add a new constructor overload to `ConfigNumberFormatter` to take a new parameter `prescale`. Use it to rescale the displayed value to a 0–100% range.

yvt (changes requested):

NotAFile:
Might it be better to have a general "ui" volume slider that applies to all such effects?

yvt (approved):

yvt:
Looks good, thanks!

PR #961, opened 2021-06-20
Increase the intensity of the bullet-block hit effect

yvt:


yvt:
The performance and gameplay-wise impact on the software renderer needs to be assessed.

PR #964, opened 2021-07-09
Player.cpp: Disable recoil if not local player

Haxk20:
This PR fixes an issue where if user with OpenSpades client is spectating
someone they would see recoil even tho there is none (Player is hacking).

This can get absolutely hard to detect and can make staff think the player
is not cheating while they absolutely are.

This leaves the cheater in game where they can make the game less
fun for others. Which is not what anybody wants.

### DIFF
--- a/Sources/Client/Player.cpp
+++ b/Sources/Client/Player.cpp
@@ -788,10 +788,12 @@ namespace spades {
 
 			horzModifier *= sqrt(1 - pow(o.z, 4));
 
-			o += GetRight() * rec.x * triWave * horzModifier;
-			o += GetUp() * std::min(rec.y, std::max(0.f, upLimit)) * vertModifier;
-			o = o.Normalize();
-			SetOrientation(o);
+			if (this->IsLocalPlayer()) {
+				o += GetRight() * rec.x * triWave * horzModifier;
+				o += GetUp() * std::min(rec.y, std::max(0.f, upLimit)) * vertModifier;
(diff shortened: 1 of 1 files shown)


Sources/Client/Player.cpp
```diff
@@ -788,10 +788,16 @@ namespace spades {
 
 			horzModifier *= sqrt(1 - pow(o.z, 4));
 
-			o += GetRight() * rec.x * triWave * horzModifier;
-			o += GetUp() * std::min(rec.y, std::max(0.f, upLimit)) * vertModifier;
-			o = o.Normalize();
-			SetOrientation(o);
+			//If the client player is spectator disable recoil
```
yvt, line 791: I think the correct distinction is whether it's a local player (for which we must apply the recoil) or a non-local player (for which we mustn't because the recoil is applied by their client, and we are supposed to just display the result). You can check that by `this->IsLocalPlayer()`.

The current code exhibits a different behavior depending on whether you are a spectator or dead and temporarily observing a teammate, which isn't probably what we want.

yvt (changes requested):

Sources/Client/Player.cpp
Haxk20: Good point. Will force push the corrected code once im home.

yvt (approved):

Haxk20:
Can this get merged since it was approved?

PR #969, opened 2021-07-25
Sources/ENet: Update ENet to latest version

Haxk20:
This commit updates the enet version from 2013 !!! to the latest commit.

Since 2013 there have been MANY improvements to the library
and we should absolutely use them since it can improve the gameplay.

This also includes the changes from previous commits as they were not upstreamed.

Tested and game runs flawesly.

### DIFF
--- a/Sources/ENet/callbacks.c
+++ b/Sources/ENet/callbacks.c
@@ -27,6 +27,12 @@ enet_initialize_with_callbacks (ENetVersion version, const ENetCallbacks * inits
 
    return enet_initialize ();
 }
+
+ENetVersion
+enet_linked_version (void)
+{
+    return ENET_VERSION;
+}
            
 void *
 enet_malloc (size_t size)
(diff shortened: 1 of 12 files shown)


Haxk20:
While i have done this i think it would be better if we upstream the changes we have in openspades to enet and then have enet as git submodule. This way updates are  very easy to do.

@yvt if you think what i said above would be the prefered way to go (It is from my opinion the far superior way to go) let me know and i will change it to submodule and even upstream the changes to enet itself.

NotAFile:
I'd be cautious about updating to latest commit vs latest version

Haxk20:
> I'd be cautious about updating to latest commit vs latest version

I have checked and the latest commit just has non code related things on top of 1.3.17 so no code changes on top of master after 1.3.17

yvt:
https://github.com/lsalzman/enet/compare/v1.3.17...74cea7abf52ddd355146aeb0a4077d2b95368122 Actually there are a few code changes, but they don't look harmful.

yvt:
@Haxk20 Can you add the exact original commit hash to the commit message?

Haxk20:
> @Haxk20 Can you add the exact original commit hash to the commit message?

yep.

PR #992, opened 2022-01-09
Add variation in SMG ejecting brass velocity

Thanson16:
Without this commit, the SMG's ejecting brass tends to fall unnaturally:
If the gun hasn't moved, each casing falls in the exact same trajectory
and in the exact same location.

This commit adds a small amount of random variation to each casing's
veloctiy, which looks more natural.

Tested on Linux, both as a player and in spectator mode.


yvt:
It looks pretty great. Thanks!

PR #993, opened 2022-01-13
cpu AVX check: check (OS)XSAVE as well.

devnexen:


yvt:
Reading <https://github.com/rust-lang/rust/blob/eac3c7c5bd38ec38062ebde475bd2ea6317d0c09/src/tools/rustfmt/tests/source/cfg_if/detect/os/x86.rs#L159> and <https://stackoverflow.com/questions/44144763/avx-feature-detection-using-sigill-versus-cpu-probing>, it seems you also need to check `XCR0.AVX[2]` to detect whether the OS supports saving and restoring AVX state by XSAVE.

yvt:
The CI failure is #994 and unrelated to this change.

yvt:
Thanks!

PR #997, opened 2022-02-13
solaris based systems port.

devnexen:
V-Sync operation unsupported by the driver and pretty rudimentary grep.
mostly linkage fixes otherwise.


yvt:
Thanks!

PR #1003, opened 2022-03-18
openal driver update proposal.

devnexen:
if we cannot have proper handle on the default, attempt
to go through to get first functionning device.

### DIFF
--- a/Resources/Scripts/Gui/StartupScreen/ConfigViewTabs.as
+++ b/Resources/Scripts/Gui/StartupScreen/ConfigViewTabs.as
@@ -473,8 +473,11 @@ namespace spades {
         StartupScreenConfigView @configViewOpenAL;
         StartupScreenConfigView @configViewYSR;
 
+        StartupScreenAudioOpenALEditor @editOpenAL;
+
         private ConfigItem s_audioDriver("s_audioDriver");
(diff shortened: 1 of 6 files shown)


Resources/Scripts/Gui/StartupScreen/ConfigViewTabs.as
```diff
@@ -467,15 +467,18 @@ namespace spades {
...
         private ConfigItem s_eax("s_eax");
+        private ConfigItem s_openalDevice("s_openalDevice");
 
         StartupScreenAudioTab(StartupScreenUI @ui, Vector2 size) {
             super(ui.manager);
             @this.ui = ui;
             @helper = ui.helper;
 
-            float mainWidth = size.x - 250.f;
+            float mainWidth = size.x - 120.f;
```
yvt, line 481: This narrow width of the help pane causes word-wrap fallback in some Slavic languages. Does the original width cause any functional issues other than long device names running over the combo box's width?

devnexen: Not at all, I did not think of other speaking languages indeed.

yvt:
Thanks!

PR #1010, opened 2022-04-16
V-Sync new submenu proposal.

devnexen:
disabled or plain mode, maybe adaptive mode can be considered at some point.


yvt:
Thanks!

PR #1043, opened 2022-09-09
feat(scripts/gui): allow selecting adaptive v-sync

devnexen:


Resources/Scripts/Gui/StartupScreen/ConfigViewTabs.as
```diff
@@ -131,11 +131,10 @@ namespace spades {
...
                         "Uses a number representation which allows wider dynamic range during rendering process. " "Additionally, this allows color calculation whose value is in linear correspondence with actual energy, " "that is, physically accurate blending can be achieved.")));
 
-                cfg.AddRow(StartupScreenConfigCheckItemEditor(
-                    ui, StartupScreenConfig(ui, "r_vsync"), "0", "1",
-                    _Tr("StartupScreen", "V-Sync"),
+                cfg.AddRow(StartupScreenConfigSelectItemEditor(
+                    ui, StartupScreenConfig(ui, "r_vsync"), "0|1|-1",
                     _Tr("StartupScreen",
-                        "Enables frame rate synchronization.")));
+                        "V-Sync|" "Off|" "Enabled|" "Smooth")));
```
yvt, line 137:
```suggestion
                        "Vertical Sync:Synchronizes screen updates to a monitor's refresh rate.|" "Off|On|Adaptive")));
```

An example from another game:
Vertical Sync: Off, On, Adaptive

Sources/Gui/SDLRunner.cpp
```diff
@@ -45,7 +45,11 @@
 SPADES_SETTING(r_videoWidth);
 SPADES_SETTING(r_videoHeight);
 DEFINE_SPADES_SETTING(r_fullscreen, "0");
+#ifdef __linux__
+DEFINE_SPADES_SETTING(r_vsync, "-1");
+#else
```
yvt, line 50: Have you verified the impact of defaulting `r_vsync` to `-1`? The failure of `SDL_GL_SetSwapInterval` is fatal, so I would rather play safe with this.

Sources/Gui/SDLRunner.cpp
```diff
@@ -518,7 +522,8 @@ namespace spades {
 
 #ifndef __sun
 					if (rtype == RendererType::GL) {
-						if (r_vsync != 0 && SDL_GL_SetSwapInterval(r_vsync) != 0) {
+						int vsync = r_vsync;
+						if (vsync != 0 && SDL_GL_SetSwapInterval(vsync) != 0) {
```
yvt, line 526: Nice change but really belongs to a separate PR.

yvt (changes requested):

Sources/Gui/SDLRunner.cpp
devnexen: Not a nice change but it s in fact a fix, you asked earlier if I tried out the default with -1 and because I did I found out the value were incorrect so needed the explicit cast.

Resources/Scripts/Gui/StartupScreen/ConfigViewTabs.as
devnexen: you re right

Sources/Gui/SDLRunner.cpp
devnexen: yes see below.

yvt: I'm not sure if you understood me. What I meant was: Are you sure the new default value `-1` is safe to use in most Linux systems? I'd hate to see new users being puzzled by an error message saying "SDL_GL_SetSwapInterval failed".

devnexen: ah ok I only tried two distros, good point, I ll revert it then

Sources/Gui/SDLRunner.cpp
yvt: Could you revert this change as well?

yvt (approved):

yvt (approved):

PR #1046, opened 2022-10-09
display client UI menu on windows min.

devnexen:


yvt (changes requested):
This would trigger whatever action assigned to the <kbd>Escape</kbd> key, e.g.:
- quitting the application (in the main or startup screen),
- closing the command console (if visible),
- showing the in-game menu (if in-game), or
- hiding the in-game menu (if already visible).

PR #1081, opened 2023-04-14
Mac OS X to macOS

nonperforming:
Now called macOS instead of Mac OS X


yvt:
Since the string returned by `GetVersionInfo` is sent to a server as part of a `VersionSend` packet, and the server could compare it to a known set of strings, it might not be a good idea to change it.

PR #1082, opened 2023-04-14
Fix Windows build path

nonperforming:
Windows uses \ instead of /
Using / leads to an error
```
'vcpkg' is not recognized as an internal or external command,
operable program or batch file.
```


yvt (approved):

yvt:
Thanks!

PR #1085, opened 2023-05-06
ci: fix macOS

nonperforming:
macos-10 is deprecated and does not work anymore.

See https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners for available runners.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,7 +38,7 @@ jobs:
 
   build-macos:
     name: Build (macOS)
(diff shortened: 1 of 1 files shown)


nonperforming:
TODO: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/

yvt (approved):

yvt:
Thanks!

